Issue #50, opened 2016-03-09
/Users/kyle.smith/.rvm/gems/ruby-2.1.6/gems/i18n-0.7.0/lib/i18n.rb:284:in `enforce_available_locales!': :en is not a valid locale (I18n::InvalidLocale)


jamesiarmes:
<a href="https://github.com/askreet"><img src="https://avatars.githubusercontent.com/u/662824?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [askreet](https://github.com/askreet)**
_Tuesday Mar 01, 2016 at 22:36 GMT_
_Originally opened as https://github.com/acquia/cloud-moonshot/issues/108_

----

Not sure if this is related to AWS-SDK v2.2.0, or some other dependency added since 0.6.0. The workaround is:

```
I18n.enforce_available_locales = false
```

Needs investigation.

jamesiarmes:
<a href="https://github.com/janost"><img src="https://avatars.githubusercontent.com/u/11500057?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [janost](https://github.com/janost)**
_Friday Mar 04, 2016 at 11:54 GMT_

----

Not sure if this is helping, but here's a stack trace:
```
Uncaught exception: I18n::InvalidLocale: :en is not a valid locale
at: /[homedir]/.rvm/gems/ruby-2.3.0/gems/i18n-0.7.0/lib/i18n.rb:284:in `enforce_available_locales!'
/[homedir]/.rvm/gems/ruby-2.3.0/gems/i18n-0.7.0/lib/i18n.rb:151:in `translate'
/[homedir]/.rvm/gems/ruby-2.3.0/gems/ruby-duration-3.2.2/lib/duration.rb:247:in `i18n_for'
/[homedir]/.rvm/gems/ruby-2.3.0/gems/ruby-duration-3.2.2/lib/duration.rb:193:in `format'
/[homedir]/.rvm/gems/ruby-2.3.0/bundler/gems/cloud-moonshot-15075146bc37/lib/moonshot/stack_asg_printer.rb:127:in `uptime_format'
/[homedir]/.rvm/gems/ruby-2.3.0/bundler/gems/cloud-moonshot-15075146bc37/lib/moonshot/stack_asg_printer.rb:120:in `instance_row'
[...]

PR #57, opened 2016-03-10
Require ruby-duration 3.2.3+ to avoid issues:

glennpratt:
- < 3.2.2 includes in activesupport json which breaks time formatting in JSON objects.
- 3.2.2 doesn't have a default locale, leading to uncaught i18n exception.


glennpratt:
Addresses #50

jamesiarmes:
:+1:

PR #59, opened 2016-03-10
Modified stack.parameters to return a hash similar to stack.outputs.

jamesiarmes:
Addresses issue #44.


askreet:
Seems fine for now. I really want to get a better interface in the future. It's unclear to plugin developers what is from the template, from the existing stack, etc.

Issue #60, opened 2016-03-10
Error from missing build script does not use InteractiveLogger.


askreet:
[ ✓ ] [ 0m 0s ] Build script build/script.sh exited successfully!
Build command did not produce output file!

PR #61, opened 2016-03-10
Fail elegantly when build script doesn't produce an output file

thaddeus:
Addresses #60

### DIFF
--- a/lib/moonshot/build_mechanism/script.rb
+++ b/lib/moonshot/build_mechanism/script.rb
@@ -41,7 +41,7 @@ class Moonshot::BuildMechanism::Script
 
   def post_build_hook(_version)
     unless File.exist?(@output_file) # rubocop:disable GuardClause
-      raise Thor::Error, 'Build command did not produce output file!'
+      step.failure 'Build command did not produce output file!'
     end
   end
 


lib/moonshot/build_mechanism/script.rb
```diff
@@ -41,7 +41,7 @@ def build_hook(version)
 
   def post_build_hook(_version)
     unless File.exist?(@output_file) # rubocop:disable GuardClause
-      raise Thor::Error, 'Build command did not produce output file!'
+      step.failure 'Build command did not produce output file!'
```
glennpratt: Not sure this will communicate the error upstream (like raise would) allowing a non-zero exit.

thaddeus: Oops, you're right. I'm fixing the CF resources I'm using, then I'll test it out again.

glennpratt: This is how @jamesiarmes handled that.

https://github.com/acquia/moonshot/blob/master/lib/moonshot/controller.rb#L67-L78

Perhaps we want to consider how we can do that more generally with the hooks.

askreet:
This PR is now incompatible with 1.0 -- please re-open if you have a chance to bring the fix up to date.

PR #63, opened 2016-03-10
Switch default S3 storage class to STANDARD_IA.

glennpratt:
Addresses #36


askreet:
Tested locally, LGTM.

PR #75, opened 2016-03-10
Add elbs instance state to status

kaushik:
#38


askreet:
Github is hanging at mergeability check on this, no idea why. In the meantime, would you mind writing some test coverage? I think we have some existing test cases around the status output helpers.

kaushik:
It's gone now :/

PR #79, opened 2016-03-16
Fixed typo in example usage of the moonshot library docs

cpliakas:
> In this example we are going to use the resources of the sample directo This example assumes you ...

Sentence seems to stop, pull request fixes it.


askreet:
Thank you!

PR #82, opened 2016-03-25
list-builds command

ghaabor:


askreet:
This is great, it highlights a bigger issue though. We need a good interface for builds management. Would love to hear any thoughts you have on #83.

PR #88, opened 2016-04-16
Simplify the sample app by using Amazon Linux and bone-stock php56 configuration.

askreet:
This patch also reduces costs to run the sample app by using t2.nano instances in HVM mode by default. Tested in a brand new AWS account with Free Tier mode. The goal of this patch was to remove the burden for new users to install, configure and understand Packer before seeing Moonshot work. Now the only requirement is properly configuring an S3 build bucket and updating scripts. I don't see a path beyond that, but it's much lower friction.

TODO:
- [x] Update sample app documentation.
- [x] Cleanup artifacts from my testing.


sample/appspec.yml
```diff
@@ -1,8 +1,8 @@
 version: 0.0
 os: linux
 files:
-  - source: docroot
-    destination: /usr/share/nginx/docroot
+  - source: docroot/
+    destination: /var/www/html
 permissions:
   - object: /usr/share/nginx
```
askreet: Oops, looks like I have to remove this as well.

sample/bin/build.sh
```diff
@@ -8,6 +8,7 @@ mkdir $BUILD_DIR
 cp $PROJECT_DIR/appspec.yml $BUILD_DIR
 cp -r $PROJECT_DIR/bin $BUILD_DIR
 cp -r $PROJECT_DIR/docroot $BUILD_DIR
+cp -r $PROJECT_DIR/config $BUILD_DIR
```
askreet: Leftover from earlier experimentation.

sample/bin/clean.sh
```diff
@@ -1,3 +1,4 @@
 #!/bin/sh
 
+rm /etc/nginx/nginx.conf
```
askreet: Leftover... Also looks like the line below should be updated. Not sure if CodeDeploy will remove files previously added. Will test.

PR #89, opened 2016-04-16
Create the `moonshot` CLI tool.

askreet:
This patch creates a moonshot CLI so that most users of Moonshot will simply install it as a global Gem, rather than building an application which requires it. This isn't feature-compatible with `Moonshot::CLI`, so we won't be removing that right away, but new features may support Moonfile first.

@glennpratt @pdrakeweb @unn -- Looking for feedback on this PR, thanks!


glennpratt:
> simply install it as a global Gem

I would still expect to use it with bundler inside my project since the parameters and configuration in the repo are still needed. I'm of the opinion global dependencies are a trap - at least in projects that are under active development. PHP / Composer world is learning this too - [site local drush](https://pantheon.io/blog/avoiding-dependency-hell-site-local-drush).

askreet:
Even tools designed to be used with other technology stacks? My concern is
that learning and using the ruby tool chain is a turnoff for people not
using Ruby as their primary means of development. Consider the aws-cli
tool. It would be awkward to suggest that someone install venv and install
the tool into a project subdirectory. Instead it installs with global pip
on a large variety of systems.

On an related note I was considering bringing our minimum Ruby version down
to 2.0 to support modern OSX/Linux systems default system Ruby and ruby
gems.

I'll review the site local drush stuff when I'm at a PC but my gut reaction
is that suggesting composer for a Drupal developer is not as much of a
stretch as suggesting RVM and friends for a Go developer.

glennpratt:
For me, yes I would probably throw a Gemfile in even on other languages. I've seen this in [Go](https://github.com/github/hub/blob/master/Gemfile) and [Rust](https://github.com/chef/delivery-cli/blob/master/Gemfile) projects. I've done the same with node dependencies in Ruby/PHP projects.

Supporting another way is also fine - just pointing out that I probably wouldn't use it that way.

askreet:
I was thinking about this again last night. I think as written we support
both, right? I would prefer the Quick start documentation show the path of
least resistance.

pdrakeweb:
I suspect that I would also avoid using/installing this as a global gem.  There are many traps in the ruby/gem world with regard to this.  Not the least of which is that depending on your ruby/rubygems version, you may execute unexpected code if you have multiple versions of a gem installed.  That said, I'm not against it being available in this way.  Maybe the quick start documentation could show how to use the moonshot bin directly and how to use moonshot as a gem within a ruby project?

askreet:
Yeah, I'm happy to document both ways -- I'll update the quickstart in this
patch.

docs/example.md
```diff
@@ -55,10 +55,10 @@ what Moonshot expects.
 
 ### Install Moonshot and it's dependencies.
 
-This step assumes that you have [Bundler](http://bundler.io/) and a modern version of Ruby installed on your system. See Moonshot's [requirements](index.md#requirements)
+Moonshot is released as a Ruby gem, and required Ruby 2.1+.
```
pingram3030: s/required/requires/

pingram3030:
A lot of other CLI tools check places like /etc/ and $HOME/ for config files, so if Moonfile is not found in `./` or the recursive `../` then would it be desirable to check, in this order, `$HOME/.moonfile` and `/etc/moonshot/moonfile`? If moonshot can be installed globally then a sane and possibly heavily commented default could be placed in to `/etc/moonshot/`.  For consistency I feel an entirely lower case filename is preferable.

askreet:
@pingram3030 Is this true of project-specific tools? Things like Vagrant require a path-local or explicit Vagrantfile, usually, no? I think it would be odd to have a global moonshot configuration -- am I really consistently starting the same known stack? And where would the cloudformation template come from? Also `/etc`?

pingram3030:
Using /etc was probably not the best example, sorry. `irb` is a [better example](http://ruby-doc.org/stdlib-2.0.0/libdoc/irb/rdoc/IRB.html#module-IRB-label-Configuration) because it has a default that can be overridden with a file in `$CWD` or an ENV var. This is only food for thought and I have no strong preference, provided things Just Work™ :).

askreet:
So, @glennpratt, I'm thinking about reviving this patch. My goal with the moonshot CLI tool is to remove some of the burden of maintaining a subclass of Moonshot::CLI, and reduce the ceremony around using the commands.

In my subclass files, I use `require 'bundler/setup'`, which eliminates the need to run, for example, `bundle exec bin/environment thing --foo`. If we presume people will go with the Gemfile approach, which is growing on me, does that mean we're all trading `bin/environment` for `bundle exec moonshot`. Do you know a better way to do this, assuming I can't do binstubs because I have two projects using Moonshot, like so:

```
proj/
  project1/
    Gemfile
    Gemfile.lock
    Moonfile
  project2/
    Gemfile
    Gemfile.lock
    Moonfile
```

I worry about if they require two different versions. Hopefully we don't completely obliterate backwards compatibility between releases, but worth thinking about.

glennpratt:
So, I use binstubs and I don't add them to my path. I've been considering a cd override that would manipulate my PATH based on dir, but that's neither here nor there.

```
binstubs/moonshot
bundle exec moonshot
PATH=binstubs/moonshot moonshot
```

All of these can run equivalent code, `binstubs/moonshot` would be similar experience to `bin/environment`.

PR #92, opened 2016-05-04
Minify stack template JSON to save bytes

pingram3030:
Per the [AWS API reference](http://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_ValidateTemplate.html) the maximum length for TemplateBody is 51,200 bytes.  If a user has a very large base template then formatting with white space takes up precious bytes.  A user of moonshot should not have to compromise the readability of their template or code their own JSON minifier.

### DIFF
--- a/lib/moonshot/stack_template.rb
+++ b/lib/moonshot/stack_template.rb
@@ -1,3 +1,5 @@
+require 'json'
+
 module Moonshot
   # A StackTemplate loads the JSON template from disk and stores information
   # about it.
@@ -18,7 +20,9 @@ module Moonshot
         raise
       end
 
-      @body = File.read(filename)
+      # The maximum TemplateBody length is 51,200 bytes, so we remove
+      # formatting white space.
+      @body = JSON.parse(File.read(filename)).to_json
     end
 
     def parameters


askreet:
Seems reasonable to me.

glennpratt:
Minor concern: does this make the stack template unreadable in the CF UI?

The alternative fix here is to use S3.

pingram3030:
It most likely does, but does that matter? This reduces many bytes worth of needless whitespace.  If a person had to rely on the JSON blob in AWS land then something has gone horribly wrong, but they could always:
```
require 'multi_json'
JSON.pretty_generate(json_blob)
```
It didn't look like moonshot could handle having the base template in S3; I may be misreading/misinterpreting though.  Regardless of this, uploading white space is a waste of electricity :)

aligentjim:
I've been evaluating Moonshot for use in a system we're currently building.  I ran into this exact situation, once I added an ELB and ASG, the CF template exceeded 51000 bytes.  

My initial attempt to work around the issue involved refactoring the various tiers of the application out into separate templates and including them as nested child stacks within the main CF template.  This worked fine (from a CloudFront perspective), but Moonshot was unable to locate the ASG resource in the template (since it was in a "child" template) when I started to try and deploy code with Moonshot.   

When I combine templates my main template and the app server tier weigh in at 55780 bytes.  Minifying the JSON gets that down to 27601 bytes.  Which makes it feasible to send via the API.  For my use case, the loss of readability in the CF console is a reasonable tradeoff for a working solution.

I like the idea of Moonshot implementing this natively per @pingram3030 PR.  Although maybe the file size should be tested, and the minification only invoked if the original size exceeds 51,000 bytes?

As @glennpratt points out, S3 is probably a better long term solution.

askreet:
It looks like this would affect the UI output, but I think we could revisit this patch with optional minifying in the future. For now I'm going to merge this as-is.

PR #95, opened 2016-05-16
Make the Deployment Configuration Name value configurable.

aligentjim:
Addes a new, optional parameter for the Deployment Configuration Name.  Defaults to "CodeDeployDefault.OneAtATime" which was originally hard coded.

Issue #93, opened 2016-05-08
Allow multiple ASGs in CodeDeploy Deployment Group


aligentjim:
Our use case for Moonshot has us deploying code to two separate ASGs (app servers and utility servers).  When I set up the CodeDeploy deployment mechanism I need to specify the logical name of a single ASG.  I've had a couple of attempts at working around this (e.g. defining my Deployment Group in the CF stack with the same name as Moonshot would have created, creating the Deployment Group via CF and not specifying an ASG at all in Moonshot).

I note there's a comment in class `Moonshot::DeploymentMechanism::CodeDeploy`:
```
  # @param app_name [String, nil] (nil)
  #   The name of the CodeDeploy Application and Deployment Group. By default,
  #   this is the same as the stack name, and probably what you want. If you
  #   have multiple deployments in a single Stack, they must have unique names.
```
Multiple deployments might be a solution to my issue, but I'm not sure how we'd implement that.  I'm not much of a Rubyist, so it's possible I'm missing something obvious in the code, in which case I'd be more than happy to update the docs with a solution if someone could point me in the right direction.

Alternatively, CodeDeploy supports multiple ASG's in a single deployment group.  Would it be better to implement support for that within Moonshot itself?  Again, happy to do the work to implement this, if we think that's the best approach.

askreet:
Is your use case wanting to deploy the same exact artifact to both types of servers? I ask because we have a team internally using Moonshot who had similar concerns, but ended up going down the path of two stacks entirely, sharing a parent "Networking" stack, because each piece of software needed separate build and deployment steps.

I've been hesitating to solve this problem because I can't think of a good one-size-fits-all solution. Some ideas are:
- Write two `Moonshot::CLI` tools which use the same application name, but specify a different ASG Logical ID and build script, which bundles a different .appspec.yml.
- Use two stacks, as our internal team did.
- Implement the concept of "Components" in moonshot where you can mix and match different mechanisms within the same stack.

The third is something I've been thinking about for a long time but each time I go to implement it, it gets increasingly complex.

aligentjim:
Our use case is to deploy exactly the same deployment artifact in exactly the same way across the two ASGs.  We have a "normal" ELB fronted ASG for our app server tier, and a second ASG which runs a "singleton" instance for cron job execution.  The same artifact is deployed to both, it's the manner of execution that differs, and that's baked into the AMI.

[...]

PR #96, opened 2016-05-16
Multiple ASGs in deployment group

aligentjim:
This PR adds the ability for Moonshot to create and validate deployment groups containing multiple ASGs per my issue #93 ...  This PR also contains the fix from my previous PR ( #95 ).

### DIFF
--- a/lib/moonshot/deployment_mechanism/code_deploy.rb
+++ b/lib/moonshot/deployment_mechanism/code_deploy.rb
@@ -229,7 +229,7 @@ class Moonshot::DeploymentMechanism::CodeDeploy # rubocop:disable ClassLength
   end
 
   def wait_for_asg_capacity
-    ilog.start 'Waiting for AutoScaling Group to reach capacity...' do |s|
+    ilog.start 'Waiting for AutoScaling Group(s) to reach capacity...' do |s|
       loop do
         asgs_at_capacity = 0
         asgs = load_auto_scaling_groups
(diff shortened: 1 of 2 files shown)


codwazny:
Hey, thanks for the contribution. Sorry for not responding earlier but I somehow had this repo silenced :( I think the code looks fine and I really appreciate you keeping it backwards compatible. I want @askreet to take a peek when he gets back from vacation in a few days.

aligentjim:
Great, thanks.  We've been using this change in the wild for a while now and it's been rock solid.  There's no reason it shouldn't be merged.

lib/moonshot/deployment_mechanism/code_deploy.rb
```diff
@@ -203,20 +225,26 @@ def create_deployment_group
       application_name: app_name,
       deployment_group_name: app_name,
       service_role_arn: role.arn,
-      auto_scaling_groups: [asg_name])
+      auto_scaling_groups: asg_names)
   end
 
   def wait_for_asg_capacity
     ilog.start 'Waiting for AutoScaling Group to reach capacity...' do |s|
```
askreet: Probably want `Autoscaling Group(s)` here as well.

askreet:
Patch looks good to me. Very sorry that this fell under our radar somehow.

lib/moonshot/deployment_mechanism/code_deploy.rb
aligentjim: Agree...  Making that change now.

PR #97, opened 2016-05-19
Introducing parameter strategies

gergelyborsothy:


gergelyborsothy:
Forgot to update docs, sorry. Will do. Feel free to take a look at the actual code meanwhile.

gergelyborsothy:
Docs updated.

docs/user-guide/stack_parameter_strategies.md
```diff
@@ -0,0 +1,35 @@
...
+and *all of them* are overridden on update.
+
+## Merge strategy
+
+Merge strategy is a new way of dealing with stack parameters on stack update.
+You only have to declare **parameters you want to update** in your parameter file,
+the remaining parameters are not updated, meaning it stays as it's current uploaded (live) state.
+This behaviour is achived by using CloudFormation's `UsePreviousValue` feature.
+This way you can avoid accidentally reverting parameter values with your outdated local
+parameter file.
```
askreet: I think once we have a command line option to send parameters this will be a really compelling case for many users simply to not have a parameters file at all. Imagine:
```
bin/environment create -n myenv -p ASGSize=4 Color=turtle
```

No parameter file is ever needed and all users can update the template json without concern for the parameters. Probably there are still cases where people would use parameter files, but I think between this and parent stacks, the use cases go way down.

docs/user-guide/stack_parameters_override.md
```diff
@@ -1,22 +0,0 @@
-# Stack Parameter overrides
```
askreet: I think there's some wisdom in here that we're losing in the new documentation. Specifically this contains examples and warnings about developers stomping on each other's settings if they are not in source control.

lib/moonshot/controller_config.rb
```diff
@@ -21,5 +25,17 @@ def initialize
       @plugins = []
       @show_all_stack_events = false
     end
+
+    def parameter_strategy=(value)
```
askreet: We talked about putting this case statement on the CLI side, I'm wondering why you ended up putting it here instead? One reason I like it on the CLI side is that it's possible for advanced users to make a new wrapper around the `Moonshot::Controller` to manage their infrastructure. If this simply accepted a class that behaved as a parameter strategy, a user could bring their own parameter strategy and run with it. With it baked into the controller config, they'd have to resort to monkeypatching this class or getting their changes merged upstream.

lib/moonshot/controller_config.rb
```diff
@@ -21,5 +25,17 @@ def initialize
...
+      @parameter_strategy =
+        case value.to_sym
+        when :default
+          Moonshot::ParameterStrategy::DefaultStrategy.new
+        when :merge
+          Moonshot::ParameterStrategy::MergeStrategy.new
+        else
+          raise Thor::Error, "Unknown parameter strategy: #{value}"
```
askreet: Assuming we keep this, we should raise a non-Thor specific error for the same reason, it's possible to use the controller without a Thor tool.

askreet:
I think this is a really good start, code is very clean. Nice work.

Have you tested this in a parent stack environment?

I'd like to see the factory logic removed from ControllerConfig, either implemented as a factory method on the ParameterStrategy module, or directly in the CLI class. Also, I want to pull this down and test it in a couple different applications before merging. I'm not sure if I can get to that today, but I will try to get to it early next week if not over the weekend.

lib/moonshot/controller_config.rb
gergelyborsothy: I'm gonna outsource it to the CLI class, but this way please note that using Moonshot without the CLI means the user not only _could_, but _**must**_ bring their own parameter strategy.

askreet: In initialize for ControllerConfig, set a new instance of the DefaultStrategy class, then you only have to override it when it's non-default and the controller is functional without CLI.

gergelyborsothy:
@askreet do you want to bump version?

askreet:
@gergelyborsothy I'll bump the version in a separate commit before cutting a release.

lib/moonshot/cli.rb
```diff
@@ -106,7 +128,11 @@ def create
     end
 
     desc :update, 'Update the CloudFormation stack within an environment.'
-    option :show_all_events, desc: 'Show all stack events during update. (Default: errors only)'
+    option :parameter_strategy, desc: 'Override default parameter strategy.'
+    option(
+      :show_all_events,
+      type: :string,
+      desc: 'Show all stack events during update. (Default: errors only)')
```
askreet: This should have been a boolean type from the beginning -- is this a bug causing undesired behavior? It certainly shouldn't be a string, though.

gergelyborsothy: No, thats a mistake, I wanted to specify the type for `parameter_strategy`, not `show_all_events`. Shall I also modify `show_all_events `'s type to boolean?

PR #100, opened 2016-06-06
Status plugins invocation fix

gergelyborsothy:
Sorry for posting this out of the blue, I just felt like this doesn't really need an issue or a previous discussion. Since documentation states there's a `pre_status` and a `post_status` method available for plugins, I fixed the controller accordingly.


askreet:
Nice catch, this will break any existing plugins that use the old method. I'll be sure to put that in the release notes.

PR #101, opened 2016-06-06
Fix bugs in the merge parameter strategy for added and dropped parame…

pdrakeweb:
…ters.

### DIFF
--- a/docs/user-guide/stack_parameter_strategies.md
+++ b/docs/user-guide/stack_parameter_strategies.md
@@ -30,7 +30,7 @@ per-environment tunings in the source repository acting as both a safety net
 and documentation of existing environments. A possible solution for using this
 strategy in a safe manner is the following:
 
-When a stack update is performed, a *parameter overrides file* is checked in
+When a stack update is performed, a *parameter file* is checked in
 `cloud_formation/parameters/environment-name.yml`. This file is YAML formatted
 and takes a hash of stack parameter names and values, for example:
 ```yaml
@@ -78,8 +78,8 @@ Example:
 
 ```ruby
 class CustomStrategy
-  def parameters(current, overrides)
-    current.map do |k, v|
+  def parameters(parameters, stack_parameters, template)
+    parameters.map do |k, v|
       {
         parameter_key: k,
         parameter_value: v,
--- a/lib/moonshot/default_strategy.rb
+++ b/lib/moonshot/default_strategy.rb
@@ -1,13 +1,12 @@
 module Moonshot
   module ParameterStrategy
-    # Default strategy: grabbing every parameter
-    # from the parameter file.
+    # Default strategy: use parameter defined in the parameter file
     class DefaultStrategy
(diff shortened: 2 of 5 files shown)


gergelyborsothy:
Cool, this behaviour was actually not expected (by me at least) for some reason, so totally makes sense. 👍

glennpratt:
👍

lib/moonshot/stack.rb
```diff
@@ -138,7 +138,7 @@ def resources_of_type(type)
     # update.
     def overrides
       if File.exist?(parameters_file)
-        YAML.load_file(parameters_file)
+        YAML.load_file(parameters_file) || {}
```
askreet: When does this return falsey? If the yaml file is "---\nfalse"?

docs/user-guide/stack_parameter_strategies.md
```diff
@@ -78,7 +78,7 @@ Example:
 
 ```ruby
 class CustomStrategy
-  def parameters(current, overrides)
+  def parameters(parameters, stack_parameters, template)
     current.map do |k, v|
```
codwazny: Current isn't defined anymore.

lib/moonshot/default_strategy.rb
```diff
@@ -1,13 +1,12 @@
...
+    # Default strategy: use parameter defined in the parameter file
     class DefaultStrategy
-      def parameters(_, overrides)
-        overrides.map do |key, _|
+      def parameters(params, _, _)
+        params.map do |key, _|
           {
             parameter_key: key,
-            parameter_value: overrides[key],
+            parameter_value: params[key],
```
codwazny: Any reason for doing this over having the map block yield value and putting value here? Pretty sure params[key] is just going to end up doing another hash look up for no reason.

lib/moonshot/stack.rb
codwazny: Also returns false if the yaml file is empty but the file exists.

lib/moonshot/default_strategy.rb
glennpratt: I agree something like that would be easier to read, though hash lookups are super fast, so probably not a big performance concern.

Could also do "safe assignment in condition":

```ruby
if (value = params[key])

codwazny: I think this comment is referring to the line 15 of merge_strategy.rb but is still relevant :)

glennpratt: Uhh, wow, yes it is.

lib/moonshot/stack.rb
pdrakeweb: If the YAML file contains valid YAML for nil ("---").

lib/moonshot/default_strategy.rb
pdrakeweb: Only because its how this is already done and I was just renaming the variable to make it more understandable.

PR #103, opened 2016-06-09
Add backup plugin

gergelyborsothy:
This is rather an initial concept. Please review with extra care.

### DIFF
--- a/docs/plugins/backup.md
+++ b/docs/plugins/backup.md
@@ -45,8 +45,8 @@ plugin(
   Backup.new do |b|
     b.bucket = 'acquia-cloud-database-test'
     b.files = [
-      { path: ['cloud_formation'], name: '%{app_name}.json' },
-      { path: ['cloud_formation', 'parameters'], name: '%{stack_name}.yml' }
+      { path: %w(cloud_formation), name: '%{app_name}.json' },
+      { path: %w(cloud_formation, parameters), name: '%{stack_name}.yml' }
     ]
     b.hooks = [:post_create, :post_update]
   end
--- a/docs/plugins/backup.md
+++ b/docs/plugins/backup.md
@@ -46,7 +46,7 @@ plugin(
     b.bucket = 'acquia-cloud-database-test'
     b.files = [
       { path: %w(cloud_formation), name: '%{app_name}.json' },
-      { path: %w(cloud_formation, parameters), name: '%{stack_name}.yml' }
+      { path: %w(cloud_formation parameters), name: '%{stack_name}.yml' }
     ]
     b.hooks = [:post_create, :post_update]
   end
--- a/docs/plugins/backup.md
+++ b/docs/plugins/backup.md
@@ -17,7 +17,6 @@ present in the same account and region as your deployment.
 When instantiating a class, you need to set the following options
 in a block, where the object is provided as a block argument:
 
-- `base_path`: the absolute path of your project directory
 - `bucket`: the name of the S3 bucket you wish to upload the tarball
 - `files`: a hash with two mandatory and one mandatory value:
   - `path` (mandatory): file path as an array, relative to the base path
@@ -31,10 +30,10 @@ in a block, where the object is provided as a block argument:
 You can use the following placeholders both in your filenames
 and tarball target names (meanings are pretty self explaining):
 
-- `<app_name>`
-- `<stack_name>`
-- `<timestamp>`
-- `<user>`
+- `%{app_name}`
+- `%{stack_name}`
+- `%{timestamp}`
+- `%{user}`
 
 ## Example
 
@@ -44,11 +43,10 @@ parameter file after create or update.
 ```ruby
 plugin(
   Backup.new do |b|
-    b.base_path = File.expand_path("..", __dir__)
     b.bucket = 'acquia-cloud-database-test'
     b.files = [
-      { path: ['cloud_formation'], name: "<app_name>.json" },
-      { path: ['cloud_formation', 'parameters'], name: "<stack_name>.yml" }
+      { path: ['cloud_formation'], name: '%{app_name}.json' },
+      { path: ['cloud_formation', 'parameters'], name: '%{stack_name}.yml' }
     ]
     b.hooks = [:post_create, :post_update]
   end
--- a/docs/plugins/backup.md
+++ b/docs/plugins/backup.md
@@ -1,0 +1,56 @@
+# Moonshot backup plugin
+
+Moonshot plugin for backing up config files.
+
+## Functionality
+
+The plugin collects and deflates certain files to a single tarball,
+and uploads that to a a given S3 bucket. The whole process happens
+in memory, nothing is written to disk. The plugin currently supports single files only, 
+including whole directories in your tarball is not possible yet.
+
+The plugin uses the Moonshot AWS config, meaning that the bucket must be
+present in the same account and region as your deployment.
+
+## Basic usage
+
+When instantiating a class, you need to set the following options
+in a block, where the object is provided as a block argument:
+
+- `base_path`: the absolute path of your project directory
+- `bucket`: the name of the S3 bucket you wish to upload the tarball
+- `files`: a hash with two mandatory and one mandatory value:
+  - `path` (mandatory): file path as an array, relative to the base path
+  - `name` (mandatory): name of the file
+  - `permission` (optional): permission in the target tarball, defaults to `0644`
+- `hooks`: which hooks to run the backup logic, works with all valid Moonshot hooks
+- `target_name`: tarball archive name, default: `<app_name>_<timestamp>_<user>.tar.gz`
+
+## Placeholders
+
+You can use the following placeholders both in your filenames
+and tarball target names (meanings are pretty self explaining):
+
+- `<app_name>`
+- `<stack_name>`
+- `<timestamp>`
+- `<user>`
+
+## Example
+
+A possible use-case is backing up a CF template and/or
+parameter file after create or update.
+
+```ruby
+plugin(
+  Backup.new do |b|
+    b.base_path = File.expand_path("..", __dir__)
+    b.bucket = 'acquia-cloud-database-test'
+    b.files = [
+      { path: ['cloud_formation'], name: "<app_name>.json" },
+      { path: ['cloud_formation', 'parameters'], name: "<stack_name>.yml" }
+    ]
+    b.hooks = [:post_create, :post_update]
+  end
+)
+```
--- a/docs/plugins/backup.md
+++ b/docs/plugins/backup.md
@@ -18,12 +18,18 @@ When instantiating a class, you need to set the following options
 in a block, where the object is provided as a block argument:
 
 - `bucket`: the name of the S3 bucket you wish to upload the tarball
-- `files`: a hash with two mandatory and one mandatory value:
-  - `path` (mandatory): file path as an array, relative to the base path
-  - `name` (mandatory): name of the file
-  - `permission` (optional): permission in the target tarball, defaults to `0644`
+- `files`: an array of relative path names as strings
 - `hooks`: which hooks to run the backup logic, works with all valid Moonshot hooks
 - `target_name`: tarball archive name, default: `<app_name>_<timestamp>_<user>.tar.gz`
+
+## Default method
+
+If you wish to back up only the current template and parameter files, you can simply
+use the factory method provided:
+
+```ruby
+plugin(Moonshot::Plugins::Backup.to_bucket('your-bucket-name'))
+```
 
 ## Placeholders
 
@@ -43,10 +49,10 @@ parameter file after create or update.
 ```ruby
 plugin(
   Backup.new do |b|
-    b.bucket = 'acquia-cloud-database-test'
+    b.bucket = 'your-bucket-name'
     b.files = [
-      { path: %w(cloud_formation), name: '%{app_name}.json' },
-      { path: %w(cloud_formation parameters), name: '%{stack_name}.yml' }
+      'cloud_formation/%{app_name}.json',
+      'cloud_formation/parameters/%{stack_name}.yml'
     ]
     b.hooks = [:post_create, :post_update]
   end
--- a/lib/moonshot.rb
+++ b/lib/moonshot.rb
@@ -14,8 +14,6 @@ module Moonshot
   end
 end
 
-require_relative 'plugins/backup'
-
 [
   # Helpers
   'creds_helper',
@@ -43,3 +41,8 @@ require_relative 'plugins/backup'
   'build_mechanism/version_proxy',
   'deployment_mechanism/code_deploy'
 ].each { |f| require_relative "moonshot/#{f}" }
+
+# Bundled plugins
+[
+  'backup'
+].each { |p| require_relative "plugins/#{p}" }
--- a/lib/moonshot.rb
+++ b/lib/moonshot.rb
@@ -10,7 +10,11 @@ module Moonshot
   end
   module DeploymentMechanism # rubocop:disable Documentation
   end
+  module Plugins # rubocop:disable Documentation
+  end
 end
+
+require_relative 'plugins/backup'
 
 [
   # Helpers
--- a/lib/plugins/backup.rb
+++ b/lib/plugins/backup.rb
@@ -33,28 +33,49 @@ module Moonshot
 
       def initialize
         yield self if block_given?
-        raise ArgumentError if @bucket.nil? || @files.nil? || @hooks.nil?
+        raise ArgumentError if @bucket.nil? || @files.nil? || @files.empty? || @hooks.nil?
+        @target_name ||= '%{app_name}_%{timestamp}_%{user}.tar.gz'
       end
 
-      def backup(resources)
+      def self.to_bucket(bucket:)
+        raise ArgumentError if bucket.nil?
+        Moonshot::Plugins::Backup.new do |b|
+          b.bucket = bucket
+          b.files = [
+            { path: %w(cloud_formation), name: '%{app_name}.json' },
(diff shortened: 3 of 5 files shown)


gergelyborsothy:
I'm not sure we need this level of flexibility/configurability, but since I had the logic already, I thought why not.
I'd also love to see the plugin optionally requirable.

lib/plugins/backup.rb
```diff
@@ -0,0 +1,117 @@
...
+        placeholder.gsub! '<stack_name>', @stack_name
+        placeholder.gsub! '<timestamp>', Time.now.to_i.to_s
+        placeholder.gsub! '<user>', ENV['USER']
+      end
+    end
+
+    # Monkeypatching TarWriter for our
+    # needs to improve clarity
+    class Gem::Package::TarWriter
+      def add(name:, path:, permission: nil)
```
glennpratt: This looks like it could be a refinement:

http://ruby-doc.org/core-2.1.1/doc/syntax/refinements_rdoc.html

lib/plugins/backup.rb
```diff
@@ -0,0 +1,117 @@
...
+      def inject_base(file)
+        file[:path] = [@base_path.clone] << file[:path]
+        file
+      end
+
+      def render_placeholder(placeholder)
+        placeholder.gsub! '<app_name>', @app_name
+        placeholder.gsub! '<stack_name>', @stack_name
+        placeholder.gsub! '<timestamp>', Time.now.to_i.to_s
+        placeholder.gsub! '<user>', ENV['USER']
```
glennpratt: I think [`sprintf` or `format`](https://github.com/bbatsov/ruby-style-guide#sprintf) would be a better choice here.

```ruby
sprintf('%{first} %{second}', first: 20, second: 10)

File: docs/plugins/backup.md [hunk 3]
```diff
@@ -0,0 +1,56 @@
...
+parameter file after create or update.
+
+```ruby
+plugin(
+  Backup.new do |b|
+    b.base_path = File.expand_path("..", __dir__)
+    b.bucket = 'acquia-cloud-database-test'
+    b.files = [
+      { path: ['cloud_formation'], name: "<app_name>.json" },
+      { path: ['cloud_formation', 'parameters'], name: "<stack_name>.yml" }
```
glennpratt: Is there much advantage to this over `'relative/path/to/file'`? I know this is debatable, but I have prefer using simple strings for paths and converting them to `Pathname` if I need some magic.

Also, moonshot should already know these path. It would be nice if it could back them up without configuring the actual path.

docs/plugins/backup.md
```diff
@@ -0,0 +1,56 @@
...
+Moonshot plugin for backing up config files.
+
+## Functionality
+
+The plugin collects and deflates certain files to a single tarball,
+and uploads that to a a given S3 bucket. The whole process happens
+in memory, nothing is written to disk. The plugin currently supports single files only, 
+including whole directories in your tarball is not possible yet.
+
+The plugin uses the Moonshot AWS config, meaning that the bucket must be
```
askreet: A bit of a nitpick, but this isn't quite true. It means that whatever credentials you are using locally must have access to the bucket. Cross-account permissions are a possibility.

docs/plugins/backup.md
gergelyborsothy: Not anymore. I was preparing this would be a separate plugin as a gem, located somewhere completely else. I think I'm actually gonna remove `base_path`, and just use relatives.

docs/plugins/backup.md
gergelyborsothy: Yeah well ¯\_(ツ)_/¯ you're right, I'm not really sure what to do about this (probably just leave it as it is?).

gergelyborsothy:
I will squash before merging.

docs/plugins/backup.md
gergelyborsothy: @glennpratt: removed base path, but unfortunately I need path and name separately, since I have to tell the TarWriter what the target filename has to be:

https://github.com/acquia/moonshot/pull/103/files#diff-6cee918ef52228fbe94d38a04372ff42R13

lib/moonshot.rb
```diff
@@ -10,8 +10,12 @@ module BuildMechanism # rubocop:disable Documentation
   end
   module DeploymentMechanism # rubocop:disable Documentation
   end
+  module Plugins # rubocop:disable Documentation
+  end
 end
 
+require_relative 'plugins/backup'
```
gergelyborsothy: Shall we use `autoload` to only load the plugin when used?

lib/plugins/backup.rb
```diff
@@ -0,0 +1,125 @@
...
+      rel_path = File.join(path << name)
+      permission ||= 0644
+
+      add_file(name, permission) do |io|
+        File.open(rel_path, 'r') { |f| io.write(f.read) }
+      end
+    end
+  end
+end
+
```
gergelyborsothy: @glennpratt: I went from monkey patching to refining, but the only way I could make this work was putting the refinement to the beginning, then say `using Moonshot::Plugins`. is this allright?

glennpratt: Yeah, I think that's how refinements work.

lib/moonshot.rb
glennpratt: I don't think we should concern ourselves with that yet. There are many files that not every user uses.

If we want to autoload because of slow startup then we should address that all at once.

glennpratt:
👍

gergelyborsothy:
@askreet let me know if you're good with this or not. I'm thinking about adding some feedback to the user that the upload is taking place, but most use-cases should be simple enough that the upload finishes in max 1-2 sec.

docs/plugins/backup.md
```diff
@@ -0,0 +1,54 @@
...
+- `%{user}`
+
+## Example
+
+A possible use-case is backing up a CF template and/or
+parameter file after create or update.
+
+```ruby
+plugin(
+  Backup.new do |b|
```
askreet: I have to imagine 95% of users of this plugin would want to backup both the json and yml contents to a target bucket. Maybe define a factory method like `Backup.to_bucket(bucket_name)` to provide a simple way to do this? We could still support the constructor w/ full config object for custom cases.

lib/plugins/backup.rb
```diff
@@ -0,0 +1,124 @@
+require 'rubygems/package'
+require 'zlib'
+require_relative '../moonshot/creds_helper'
+
+module Moonshot::Plugins # rubocop:disable Documentation
+  # Monkeypatching TarWriter for our
```
askreet: "for our needs" is vague, can we spell out what the original `TarWriter#add` implementation does that is undesirable and how we change it?

mkdocs.yml
```diff
@@ -13,6 +13,8 @@ pages:
   - 'Build': 'mechanisms/build.md'
   - 'Deployment': 'mechanisms/deployment.md'
 - Plugins: plugins.md
+- Bundled plugins:
```
askreet: On line 17 the string with a space uses single quotes, but this one does not. Is it required?

lib/plugins/backup.rb
```diff
@@ -0,0 +1,124 @@
...
+        raise e
+      ensure
+        tar_out.close unless tar_out.nil?
+        zip_out.close unless zip_out.nil?
+      end
+
+      # dynamically responding to hooks
+      # supplied in the constructor
+      def method_missing(method_name, *args, &block)
+        backup(*args) if @hooks.include?(method_name) || super
```
askreet: I think the precedence here might not be what you expect. For example, if `@hooks.include?(method_name)` is false, super will be called. If super returns truthy, `backup(*args)` will be called. I think what you want here is more like:
```ruby
if @hooks.include?(method_name)
  backup(*args)
else
  super
end
```

As an example:
```ruby
def do_thing
	puts "I did a thing!"
end

def have_thing
	false
end

def other_thing
	puts "I did another thing!"
	true
end

do_thing if have_thing || other_thing
```

This code produces:
```
I did another thing!
I did a thing!
=> nil
```

I think you expected it to produce only `I did another thing!`? Of course if the super returned falsey, that would be all that was produced.

In my experience, it's best to be very clear and verbose with code. Optimize for humans :-).

docs/plugins/backup.md
askreet: I might just omit this line if I were you. I don't know if we added doctor support for plugins, but if we did (or you felt like adding it) this plugin could *check* if a user has read/write access to the bucket when you run `bin/<tool> doctor`.

askreet:
Conceptually I think this is very cool and I could imagine many users getting benefit from it.

Things I'd like to see before commit:
- Unit tests.
- Some cleanup of the method documentation.
- Feedback to the user (as you said above) using the `InteractiveLogger` provided by `Resources.`

docs/plugins/backup.md
gergelyborsothy: I like the idea, although `to_bucket` can be a bit misleading in a way that this plugin is only capable of uploading the tarball to an S3 bucket.

lib/plugins/backup.rb
gergelyborsothy: Truth be told, `add` isnt a method in TarWriter, it just makes the implementation way easier if I patch it to the class.

lib/plugins/backup.rb
gergelyborsothy: Fixed.

mkdocs.yml
gergelyborsothy: Based on the patterns on the lines above, I think no.

docs/plugins/backup.md
askreet: I could not come up with a name I liked for the factory method :-(

lib/plugins/backup.rb
```diff
@@ -3,8 +3,8 @@
 require_relative '../moonshot/creds_helper'
 
 module Moonshot::Plugins # rubocop:disable Documentation
-  # Monkeypatching TarWriter for our
-  # needs to improve clarity
+  # Defining new method for the ease of use in this
```
askreet: This doesn't explain what "this particular use case is" -- what does the original method do and why are you changing it? That documentation will definitely save someone time in the future.

lib/plugins/backup.rb
```diff
@@ -33,28 +33,57 @@ class Backup
...
-        zip_out = zip tar_out
-        upload zip_out
-      rescue StandardError => e
-        raise e
-      ensure
-        tar_out.close unless tar_out.nil?
-        zip_out.close unless zip_out.nil?
+        @target_name = render(@target_name)
+
+        resources.ilog.start log_message.clone << ' in progress.' do |s|
```
askreet: nit: Interpolation would accomplish the same thing and is easier to grok, imo.
```ruby
resources.ilog.start("#{log_message} in progress.") do |s|
   # ...
end
```

lib/plugins/backup.rb
```diff
@@ -33,28 +33,57 @@ class Backup
...
+          begin
+            tar_out = tar @files
```
askreet: We generally use parenthesis for method invocation in this codebase, best to be consistent in this method.

lib/plugins/backup.rb
```diff
@@ -111,6 +148,9 @@ def render(placeholder)
         )
       end
 
+      # Renders string with the specified placeholders
+      # @param file [String] array of raw strings
```
askreet: You should list container types directly, and include an example:
```
# @param file [Array<String>]
#   A list of file paths (e.g. 'an example here'.
```

docs/plugins/backup.md
```diff
@@ -46,7 +46,7 @@ plugin(
     b.bucket = 'acquia-cloud-database-test'
     b.files = [
```
askreet: Is there a reason not to make files simply:
```ruby
b.files = [
  'cloud_formation/%{app_name}.json',
  'cloud_formation/parameters/%{stack_name}.yml'
]
```

lib/plugins/backup.rb
```diff
@@ -66,8 +95,9 @@ def respond_to?(method_name, include_private = false)
       attr_accessor :app_name,
                     :stack_name
 
-      # returns a tar IO object
-      # from the passed files
+      # Creates an IO stream from the passed files
```
askreet: Can I suggest something like:
```ruby
# Create a tar archive in memory, returning the IO object pointing at the beginning of the archive.
#
# @param target_files [Array<String>]
# @return tar_stream [IO]
```

lib/plugins/backup.rb
gergelyborsothy: Absolutely 👍

lib/plugins/backup.rb
```diff
@@ -92,15 +123,21 @@ def zip(io_tar)
         zip_stream
       end
 
+      # Uploads an object from the pass
```
askreet: *passed -- what column are we wrapping these comments to? I think 80 is a good standard.

lib/plugins/backup.rb
gergelyborsothy: Fixed.

spec/moonshot/plugins/backup_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+      :pre_doctor,
+      :post_doctor
+    ]
+  end
+
+  describe '#new' do
+    subject { Moonshot::Plugins::Backup }
+
+    it 'should yield self' do
+      expect do |b|
```
askreet: I don't think this is right. You can test the behavior like so:
```ruby
c = nil
b = subject.new do |c|
  c.bucket = 'test'
end
expect(b).to be(c)
expect(b.bucket).to eq('test')
```

docs/plugins/backup.md
gergelyborsothy: I need the target filename for the tarball, which I either retrieve separately (as of now, in a hash), or go with some string chopping magic. You can also optionally specify the target permissions.
https://github.com/acquia/moonshot/pull/103/files#diff-6cee918ef52228fbe94d38a04372ff42R9

lib/plugins/backup.rb
gergelyborsothy: _what does the original method do_ --> there is no original method, I can further elaborate why I thought this is useful tho.

lib/plugins/backup.rb
gergelyborsothy: Refactored comments to wrap at 80.

spec/moonshot/plugins/backup_spec.rb
gergelyborsothy: This also makes sense, what I was trying to achieve here was to inspect whether it yields at all or not.

lib/plugins/backup.rb
askreet: If there is no original method, why monkey-patch at all? Can you just use a private method in our own namespace?

docs/plugins/backup.md
askreet: Just use `File.basename` -- it's not magic, it's just how filesystems work.

```ruby
2.1.6 :002 > File.basename("/path/to/file.txt")
 => "file.txt"
2.1.6 :004 > File.basename("/path/to\n\a\ndirectory/a file\\with\nweirdcharacters.dat")
 => "a file\\with\nweirdcharacters.dat"
```

spec/moonshot/plugins/backup_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+    it 'should raise ArgumentError if resources not injected' do
+      expect { subject.backup }.to raise_error(ArgumentError)
+    end
+
+    it 'should retrieve values from resources' do
+      expect(resources).to receive(:stack)
+      expect(resources.stack).to receive(:app_name)
+      expect(resources.stack).to receive(:name)
+      expect(resources.ilog).to receive(:start)
+      subject.backup(resources)
```
askreet: Does this work? Seems like this would call out to S3, no?

We probably want to push this kind of testing into an integration suite. Unfortunately, we don't have one yet so I think we can pass on this level of testing in `spec/`.

lib/plugins/backup.rb
```diff
@@ -0,0 +1,168 @@
...
+        format(
+          placeholder,
+          app_name: @app_name,
+          stack_name: @stack_name,
+          timestamp: Time.now.to_i.to_s,
+          user: ENV['USER']
+        )
+      end
+
+      # Renders string with the specified placeholders
```
askreet: I think if you switch to an array of paths, this method just goes away and becomes something like:
```ruby
paths = paths.map { |p| render(p) }
```

spec/moonshot/plugins/backup_spec.rb
gergelyborsothy: For some reason it does work, I think the S3 client might silence exceptions? Either way, I removed it.

docs/plugins/backup.md
gergelyborsothy: Allright, shall I just drop the option to set permissions?

askreet: Yeah, my preference is always for *simple*. I can't think of a use case for asserting the permissions within the tarball.

lib/plugins/backup.rb
gergelyborsothy: Ah yeah absolutely, I dont even know why I went that way. Fixed.

docs/plugins/backup.md
gergelyborsothy: Fixed. I'm not really sure what this has to do with how filesystems work, although it's true that I wasn't aware of the existence of that function.

lib/plugins/backup.rb
gergelyborsothy: I'm sure I had a great reason, the code has evolved so much ever since that I don't know what it was. Patch removed.

PR #107, opened 2016-06-22
Backup plugin buckets enhancement

gergelyborsothy:


lib/plugins/backup.rb
```diff
@@ -45,6 +48,9 @@ def backup(resources) # rubocop:disable Metrics/AbcSize
         @app_name = resources.stack.app_name
         @stack_name = resources.stack.name
         @target_name = render(@target_name)
+        @target_bucket = define_bucket
+
+        return if @target_bucket.nil?
```
gergelyborsothy: @askreet at this point, a single bucket string or a hash map is definitely provided, but it's also possible that the hash map doesn't contain the current account as a key. I'd suggest to silence the plugin in this scenario, eg. we don't want to back up the archives during system tests.

lib/plugins/backup.rb
```diff
@@ -138,7 +145,29 @@ def render(placeholder)
...
+      end
+
+      def iam_account
+        iam_client.list_account_aliases.account_aliases.first
+      end
+
+      def define_bucket
+        case
+        # returning already calculated bucket name
+        when !@target_bucket.nil?
```
askreet: No need for double-negative here:
```ruby
when @target_bucket
  @target_bucket
```

gergelyborsothy: Good call. Fixed

lib/plugins/backup.rb
```diff
@@ -138,7 +145,29 @@ def render(placeholder)
...
+        when !@bucket.nil?
+          @bucket
+        # calculating bucket based on account name
+        when !@buckets.nil?
+          bucket_by_account(iam_account)
+        end
+      end
+
+      def bucket_by_account(account)
+        @buckets[account] if @buckets.include?(account)
```
gergelyborsothy: Fixed.

lib/plugins/backup.rb
askreet: Okay, that seems fine.

lib/plugins/backup.rb
```diff
@@ -138,7 +145,29 @@ def render(placeholder)
       end
 
       def log_message
-        "Uploading '#{@target_name}' to '#{@bucket}'"
+        "Uploading '#{@target_name}' to '#{@target_bucket}'"
+      end
+
+      def iam_account
+        iam_client.list_account_aliases.account_aliases.first
```
askreet: What happens in an account without aliases? I wonder what we want the behavior to be.

spec/moonshot/plugins/backup_spec.rb
```diff
@@ -32,6 +32,17 @@
...
+        subject.new do |b|
+          b.bucket = 'test'
+          b.buckets = {}
+          b.files = %w(sample files)
+          b.hooks = [:sample, :hooks]
+        end
+      end.to raise_error(ArgumentError)
+    end
+
     let(:backup) do
```
askreet: Do we want to add better test coverage of the bucket behavior in the hash case, e.g. if no account alias is set or no account alias matches it would be nice to assert this behavior.

askreet:
Code looks good, I think there is more we can do on the RSpec side to cover the intended behaviors, though.

lib/plugins/backup.rb
askreet: I tested and this value is nil. So we would silently not backup, right?

gergelyborsothy: Yes, that's correct.

PR #108, opened 2016-07-05
CPD-885 Grant access to S3 artifacts created in other accounts.

glennpratt:
Previously, files uploaded by other accounts were not accessible by the bucket owner.

For example, if we created deployments in our "production" account, they were unreadable in our "development" account.


jamesiarmes:
:+1:

Issue #72, opened 2016-03-10
Create an ssh command


jamesiarmes:
<a href="https://github.com/glennpratt"><img src="https://avatars.githubusercontent.com/u/202230?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [glennpratt](https://github.com/glennpratt)**
_Wednesday Mar 02, 2016 at 21:42 GMT_
_Originally opened as https://github.com/acquia/cloud-moonshot/issues/110_

----

Create a command to help ssh to a server in an ASG (interactive, maybe even parallel). Maybe some hooks for deployment dir or other things.

askreet:
This was implemented back in #110. Woo!

PR #110, opened 2016-07-10
SSH functionality

janost:
Implemented most of the features requested in #72.
By default the `ssh` command selects a random instance from the stack and opens an interactive SSH session. The instance ID, username, private key file and the command can be specified in the command line.

The `#ssh_command` and the `#ssh_options!` methods in `Moonshot::CLI` are there because with the help of them a user can implement custom SSH commands in their CLI tools pretty easily.
Example usage:
```ruby
desc :'sql-cli', 'Jump into mysql console on an instance.'
ssh_options!
def sql_cli
  ssh_command('mysql -u blah')
end
```

### DIFF
--- a/lib/moonshot/cli.rb
+++ b/lib/moonshot/cli.rb
@@ -47,6 +47,24 @@ module Moonshot
         base.include(Moonshot::BuildMechanism)
         base.include(Moonshot::DeploymentMechanism)
       end
+
+      def ssh_options!
+        option :user, default: ENV['MOONSHOT_SSH_USER'] || ENV['USER'], type: :string,
+                      aliases: '-l',
+                      desc: 'Specifies the user to log in as on the remote machine.'
+        option :identity_file, default: ENV['MOONSHOT_SSH_KEY_FILE'], type: :string,
+                               aliases: '-i',
+                               desc: 'Selects a file from which the identity (private key) for '\
+                                     'public key authentication is read.'
+        option :instance, default: nil, type: :string, aliases: '-s',
+                          desc: 'Connect to specified instance ID instead of the first one.'
+        option :command, default: nil, type: :string, aliases: '-c',
+                         desc: 'Execute the specified command instead of opening a shell.'
+        option :auto_scaling_group, default: nil, type: :string, aliases: '-g',
+                                    desc: 'The logical ID of the auto scaling group to SSH into. '\
+                                    'Mandatory if the stack has multiple ASGs, '\
+                                    'ignored if you have only one.'
+      end
     end
 
     def initialize(*args)
@@ -91,6 +109,12 @@ module Moonshot
           parameter_strategy = options[:parameter_strategy] || self.class.default_parameter_strategy
           config.parameter_strategy = parameter_strategy_factory(parameter_strategy) \
             unless parameter_strategy.nil?
+
+          config.ssh_user = options[:user]
+          config.ssh_identity_file = options[:identity_file]
+          config.ssh_instance = options[:instance]
+          config.ssh_command = options[:command]
+          config.ssh_auto_scaling_group_name = options[:auto_scaling_group]
         end
       rescue => e
         raise Thor::Error, e.message
@@ -105,6 +129,12 @@ module Moonshot
         else
           raise Thor::Error, "Unknown parameter strategy: #{value}"
         end
+      end
+
+      def ssh_command(command)
+        arguments = ARGV.drop(1)
+        arguments += ['--command', command]
+        invoke :ssh, [], arguments
       end
     end
 
@@ -179,6 +209,12 @@ module Moonshot
     def doctor
       success = controller.doctor
       raise Thor::Error, 'One or more checks failed.' unless success
+    end
+
+    desc :ssh, 'SSH into the first or specified instance on the stack.'
+    ssh_options!
+    def ssh
+      controller.ssh
     end
   end
 end
--- a/lib/moonshot/controller.rb
+++ b/lib/moonshot/controller.rb
@@ -80,6 +80,12 @@ module Moonshot
       success
     end
 
+    def ssh
+      run_plugins(:pre_ssh)
+      stack.ssh
+      run_plugins(:post_ssh)
+    end
+
     def stack
       @stack ||= Stack.new(stack_name,
                            app_name: @config.app_name,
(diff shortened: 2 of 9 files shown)


askreet:
Love this! A few high level things before I dive into the code and try it out:

Do we want to randomly choose any instance from any auto-scaling group? What if a stack has two tiers and one tier has significantly different access than another? I think instead we want to default to choosing the only ASG if only one ASG exists (a common use case, for sure), otherwise throwing an error requiring the Moonshot::ControllerConfig to have a "ssh_auto_scaling_group_name" option set that dictates which ASG to find an instance in. If we want to get this in in phases (in the interest of time), we could implement an error message when used against stacks with more than one ASG for now, and implement the config option in the future.

On the topic of "random", I also think we want to do something predictable like use the lowest isntance ID when sorted alphabettically. Reason being I could see a use case where two developers are debugging a problem and one developer logs in to gather some debug data and leaves it on the instance. Dev 1 could say to dev 2, "Just log in with bin/environment ssh and take a look at my debug output", for example.

Thanks again for putting the patch together, this is a really great feature that a few teams I'm aware of (including mine) are implementing manually. I'll look through the code later today or tomorrow.

lib/moonshot/cli.rb
```diff
@@ -47,6 +47,21 @@ def inherited(base)
         base.include(Moonshot::BuildMechanism)
         base.include(Moonshot::DeploymentMechanism)
       end
+
+      def ssh_options!
+        option :user, default: ENV['MOONSHOT_SSH_USER'] || ENV['USER'], type: :string,
+                      aliases: '-l',
+                      desc: 'Specifies the user to log in as on the remote machine.'
+        option :'identity-file', default: ENV['MOONSHOT_SSH_KEY_FILE'], type: :string,
```
askreet: If you use `:identity_file`, Thor will handle both `--identity_file` and `--identity-file` correctly, and document the hyphenated form in `--help`.

lib/moonshot/controller_config.rb
```diff
@@ -15,6 +15,10 @@ class ControllerConfig
     attr_accessor :plugins
     attr_accessor :show_all_stack_events
     attr_accessor :parameter_strategy
+    attr_accessor :instance
```
askreet: I'd like to see these called `ssh_*`.

lib/moonshot/stack.rb
```diff
@@ -93,6 +93,18 @@ def status
...
+    def ssh
+      box_id = @config.instance || instances.sample
+      box_ip = instance_ip(box_id)
+      cmd = ['ssh', '-t']
+      cmd << "-i #{@config.identity_file}" if @config.identity_file
+      cmd << "-l #{@config.user}" if @config.user
+      cmd << box_ip
+      cmd << @config.command if @config.command
+      puts "Opening SSH connection to #{box_id} (#{box_ip})..."
+      system(cmd.join(' '))
```
askreet: In our command we use `exec` instead of `system`. It probably saves some minuscule amount of memory, but seems better since we don't need the use the result of the command for anything.

lib/moonshot/stack.rb
```diff
@@ -176,6 +188,25 @@ def add_parameter_overrides(hash)
 
     private
 
+    def asgs
+      resources_of_type('AWS::AutoScaling::AutoScalingGroup')
+    end
+
+    def instance_ip(instance_id)
+      Aws::EC2::Client.new.describe_instances(instance_ids: [instance_id])
```
askreet: This has the potential to raise a variety of non-intuitive errors. For example:
- What if EC2 and AutoScaling don't agree on the state of the instance and Reservations[].Instances[] comes back empty?
- What if the instance is still booting and doesn't have a public IP address?

I think as a first pass we could just rescue StandardError (i.e. naked `rescue`) and re-raise a RuntimeError that says "Failed to determine Instance IP."

lib/moonshot/stack.rb
```diff
@@ -176,6 +188,25 @@ def add_parameter_overrides(hash)
...
+                      .reservations.first.instances.first.public_ip_address
+    end
+
+    def instances
+      result = []
+      Aws::AutoScaling::Client.new.describe_auto_scaling_groups(
+        auto_scaling_group_names: asgs.map(&:physical_resource_id)
+      ).auto_scaling_groups.each do |asg|
+        result += asg.instances.map(&:instance_id)
```
askreet: I would just stick with the functional style in this method:
```ruby
def instances
  client = Aws::AutoScaling::Client.new
  client.describe_auto_scaling_groups(auto_scaling_group_names: asgs.map(&:physical_resource_id))
    .map { |asg| asg.instances.map(&:instance_id) }
    .flatten
end
```

That being said if we go with the "you must specify the ASG if you have more than one" suggestion I made in the PR, this method would become much simpler anyway.

askreet:
Code looks good. Let me know what you think about my earlier suggestions, then I'll take it for a spin against our project as a test. Thanks again, really happy to see this get into core. :)

lib/moonshot/cli.rb
janost: I didn't know that, thanks.

janost:
Thanks for the suggestions and comments @askreet, I'll review them and think about them soon.

lib/moonshot/controller.rb
```diff
@@ -80,6 +80,12 @@ def doctor
       success
     end
 
+    def ssh
+      run_plugins(:pre_ssh)
+      stack.ssh
+      run_plugins(:pre_ssh)
```
askreet: This should be post_ssh.

janost: Oops.

janost:
Based on @askreet's comments I've changed the following:
- Instead of connecting to a random instance, we are connecting to the first one alphabetically, unless the user specifies an instance ID.
- Implemented support for multiple ASGs on the stack. If there's only one, we are automatically looking for instances in that. In case of multiple ASGs, the user must specify the ASG's Logical ID using the `--auto-scaling-group` command line option.
- Small fixes/changes.

askreet:
@janost I just pulled this down and tested against our stack. I was able to delete 30 lines of code from our subclass :D. Is there anything else you want to change before I merge? I think it's ready to go.

janost:
Thanks @askreet, I think I've implemented everything I originally planned.

janost:
Squashed my commits into one.

PR #114, opened 2016-07-27
Added plugin to remove artifacts after a stage has been deleted.

Skarlso:
Purpose:

- Too many artifacts litter an s3 bucket. They need to be removed with the stage.

### DIFF
--- a/lib/moonshot.rb
+++ b/lib/moonshot.rb
@@ -43,6 +43,8 @@ end
 ].each { |f| require_relative "moonshot/#{f}" }
 
 # Bundled plugins
+# rubocop:disable Style/WordArray
 [
-  'backup'
+  'backup',
+  'cleaner'
 ].each { |p| require_relative "plugins/#{p}" }
--- a/lib/plugins/cleaner.rb
+++ b/lib/plugins/cleaner.rb
@@ -1,0 +1,47 @@
+require 'rubygems/package'
+require_relative '../moonshot/creds_helper'
+require_relative '../moonshot/resources_helper'
+
+module Moonshot
+  module Plugins
+    # Moonshot plugin class for removing artifacts when a stage gets deleted.
+    class Cleaner
+      include Moonshot::CredsHelper
+      include Moonshot::ResourcesHelper
+
+      AWS_OBJECT_LIMIT = 1000
+
+      def initialize(bucket_name)
+        raise ArgumentError if bucket_name.empty?
+        @bucket_name = bucket_name
+      end
+
+      def post_delete(resources)
+        @resources = resources
+        ilog.start "Starting to delete artifacts for stack: #{stack.name}" do |s|
+          arts = retrieve_artifact_names(stack.name)
+          if arts.empty?
+            s.success 'No artifacts found, nothing to delete.'
+          else
+            delete_artifacts(arts)
+            s.success "Deleted artifacts for stack: '#{stack.name}' in bucket: '#{@bucket_name}'"
+          end
+        end
+      end
+
+      def retrieve_artifact_names(name)
+        objs = s3_client.list_objects(bucket: @bucket_name, prefix: name)
+        objs.contents.map(&:key).select { |k| k =~ /^#{name}(\-\d+)?\.(tar\.gz|zip|tar)$/ }
+      end
+
+      def delete_artifacts(artifacts)
+        objects = artifacts.map { |a| { key: a } }
+        objects.each_slice(AWS_OBJECT_LIMIT) do |s|
+          resp = s3_client.delete_objects(bucket: @bucket_name,
+                                          delete: { objects: s })
+          ilog.error "Could not delete artifacts. Error: #{resp.errors}" unless resp.errors.empty?
+        end
+      end
+    end
+  end
+end
--- a/spec/moonshot/plugins/cleaner_spec.rb
+++ b/spec/moonshot/plugins/cleaner_spec.rb
@@ -1,0 +1,97 @@
+describe Moonshot::Plugins::Cleaner do
+  include_context 'with a working moonshot application'
+  let(:resources) do
+    instance_double(
+      Moonshot::Resources,
+      stack: instance_double(Moonshot::Stack, app_name: 'test_app_name', name: 'test_name'),
+      ilog: instance_double(Moonshot::InteractiveLoggerProxy)
+    )
+  end
+
+  before(:all) do
+    # Using doubles without context is unsupported. with_temporary_scope to the rescue.
+    RSpec::Mocks.with_temporary_scope do
+      allow(Aws::S3::Client).to receive(:new).and_return(instance_double(Aws::S3::Client))
+    end
+  end
+
+  subject do
+    described_class.new('bucket')
+  end
+
+  describe '#new' do
+    it 'should raise ArgumentError if insufficient parameters are provided' do
+      expect { described_class.new }.to raise_error(ArgumentError)
+    end
+  end
+
+  describe '#list_objects' do
+    it 'should return with the correct names if a search prefix is provided' do
+      contents = [{ key: 'file1.tar.gz' }, { key: 'file2.tar.gz' }]
+      Aws.config[:s3] = {
+        stub_responses: {
+          list_objects: { contents: contents }
+        }
+      }
+      expect(subject.retrieve_artifact_names('file1')).to eq(['file1.tar.gz'])
+    end
+
+    it 'it should not fail if no artifacts are found' do
+      contents = [{ key: 'file1.tar.gz' }, { key: 'file2.tar.gz' }]
+      Aws.config[:s3] = {
+        stub_responses: {
+          list_objects: { contents: contents }
+        }
+      }
+      expect(subject.retrieve_artifact_names('liludalas')).to eq([])
+    end
+  end
+
+  describe '#delete_artifacts' do
+    it 'should log an error message if the delete operation failed' do
+      subject.resources = resources
+      contents = [{ key: 'Error1', version_id: '1', code: '1', message: 'failed' }]
+      Aws.config[:s3] = {
+        stub_responses: {
+          delete_objects: { errors: contents }
+        }
+      }
+      expect(resources.ilog).to receive(:error)
+      subject.delete_artifacts(['liludalas'])
+    end
+
+    it 'should not log an error message in case the delete was successful' do
+      subject.resources = resources
+      contents = [{ key: 'file1', version_id: '1' }]
+      Aws.config[:s3] = {
+        stub_responses: {
+          delete_objects: { deleted: contents }
+        }
+      }
+      expect(resources.ilog).not_to receive(:error)
+      subject.delete_artifacts(['file1'])
+    end
+  end
+
+  describe '#post_delete' do
+    let(:step) { instance_double('InteractiveLogger::Step') }
+
+    it 'post_delete should inform what it deleted from where' do
+      expect(resources.ilog).to receive(:start).and_yield(step)
+      expect(step).to receive(:success)
+        .with('Deleted artifacts for stack: ' \
+              "'#{resources.stack.name}' in bucket: 'bucket'")
+      expect(subject).to receive(:retrieve_artifact_names).with(resources.stack.name)
+        .and_return(['test_name'])
+      subject.post_delete(resources)
+    end
+
+    it 'post_delete should inform if nothing was deleted' do
+      expect(resources.ilog).to receive(:start).and_yield(step)
+      expect(step).to receive(:success).with('No artifacts found, nothing to delete.')
+      expect(subject).to receive(:retrieve_artifact_names).with(resources.stack.name)
+        .and_return([])
+      subject.post_delete(resources)
+    end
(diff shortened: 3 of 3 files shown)


spec/moonshot/plugins/cleaner_spec.rb
```diff
@@ -0,0 +1,22 @@
+describe Moonshot::Plugins::Cleaner do
```
Skarlso: This is in progress to write some proper tests for the class. Please ignore this for now.

lib/plugins/cleaner.rb
```diff
@@ -0,0 +1,47 @@
...
+        end
+      end
+
+      private
+
+      def retrieve_artifact_names(name)
+        objs = s3_client.list_objects(bucket: @bucket_name, prefix: "#{name}")
+        valid_keys = []
+        objs.contents.each { |o| valid_keys << o.key if o.key =~ /^#{name}[\.-]/ }
+        raise Thor::Error,
```
codwazny: Do we really care if there are no artifacts to clean up? What if we never did a deploy. Then we will never be able to delete the stack at all, right?

lib/plugins/cleaner.rb
```diff
@@ -0,0 +1,47 @@
...
+  module Plugins
+    # Moonshot plugin class for removing artifacts when a stage gets deleted.
+    class Cleaner # rubocop:disable Metrics/ClassLength
+      include Moonshot::CredsHelper
+
+      def initialize(bucket_name)
+        @bucket_name = bucket_name
+      end
+
+      def pre_delete(resources)
```
codwazny: I wonder if pre_delete is what we want here. The artifacts are in use before we delete the stack. Perhaps post_delete is a better fit?

lib/plugins/cleaner.rb
```diff
@@ -0,0 +1,47 @@
...
+              "Artifact not found with name:" \
+              "#{name} in bucket: #{@bucket_name}!" if valid_keys.count < 1
+        valid_keys
+      end
+
+      def delete_artifacts(artifacts)
+        objects = artifacts.map { |a| { key: a } }
+        resp = s3_client.delete_objects(bucket: @bucket_name,
+                                        delete: { objects: objects,
+                                                  quiet: false } )
```
codwazny: What does quiet false do? Does it just spill a ton of crap to the terminal?

lib/plugins/cleaner.rb
Skarlso: Hah. I had this as post_delete before. I wonder when did I change it...

lib/plugins/cleaner.rb
Skarlso: Good point. Fair enough. Going to remove the error.

lib/plugins/cleaner.rb
Skarlso: No idea, whatsoever. It did not, indeed display anything. It was required though.

lib/plugins/cleaner.rb
```diff
@@ -0,0 +1,47 @@
...
+        raise Thor::Error,
```
codwazny: Being that this is currently a pre_delete step, I wonder if throwing an error here is what everyone would expect. When it comes down to cost, terminating the stack is significantly more important than the deployments in the bucket. Perhaps this is another great reason why we want this as a post_delete hook. Also, be advised that callers MAY (us included) have additional hooks running and by throwing an exception here, you kill any following hooks which might not be entirely expected.

I don't think we've really discussed how hooks should work when they failed to do their job. Should they raise an exception and kill the whole delete chain or should they just log an error message and continue? If they do log an error message, how does that affect the exit code of the Moonshot::CLI. I bet if we just log an error message line it would exit 0 which is obviously wrong still so there is clearly some work to be done here in general.

askreet: Perhaps the most important reason not to cause a failure here is that certain types of stacks will never have an associated artifact by design. The `deploy-code` action is really designed for development. CI and Production should be using named artifacts, generally, which would not be cleaned up by a teardown process.

lib/plugins/cleaner.rb
```diff
@@ -0,0 +1,51 @@
...
+    class Cleaner
+      include Moonshot::CredsHelper
+
+      def initialize(bucket_name)
+        @bucket_name = bucket_name
+      end
+
+      def post_delete(resources)
+        stack_name = resources.stack.name
+        resources.ilog.start 'Starting to delete artifacts for stack: ' \
```
askreet: We have a 100 character line limit in this project, not 80. Can this fit on 100 characters. I find this newline makes this much harder to read.

askreet: If you include `ResourcesHelper`, you get a reader method called `ilog` that could be used to reduce the length of this line without sacrificing readability as well.

lib/plugins/cleaner.rb
```diff
@@ -0,0 +1,51 @@
...
+          "#{stack_name}" do |s|
+          arts = retrieve_artifact_names(stack_name)
+          if arts.empty?
+            s.success 'No artifacts found, nothing deleted.'
+          else
+            delete_artifacts arts
```
askreet: Parenthesis here, please. :)

lib/plugins/cleaner.rb
```diff
@@ -0,0 +1,51 @@
...
+            s.success "Deleted artifacts for stack: '#{stack_name}' " \
```
askreet: Same here, also can we make this fit on a 100 character line? Might just be me but these linebreaks do a number on readability.

lib/plugins/cleaner.rb
```diff
@@ -0,0 +1,51 @@
...
+        objs.contents.each do |o|
+          valid_keys << o.key if o.key =~ /^#{name}[\.-]/
+        end
+        valid_keys
+      end
+
+      def delete_artifacts(artifacts)
+        objects = artifacts.map { |a| { key: a } }
+        resp = s3_client.delete_objects(bucket: @bucket_name,
+                                        delete: { objects: objects } )
```
askreet: DeleteObjects accepts a maximum of 1,000 items. We should do some kind of `#each_slice` operation here.

lib/plugins/cleaner.rb
```diff
@@ -0,0 +1,51 @@
...
+        raise Thor::Error,
```
askreet: The Moonshot controller shouldn't really raise Thor errors directly, it's designed to be an abstraction layer. I think the CLI classes already catch `RuntimeError` and re-raise a `Thor::Error`? If not, let's raise a `RuntimeError` here anyway and open an issue to make the CLI layer do the right thing.

lib/plugins/cleaner.rb
```diff
@@ -0,0 +1,51 @@
...
+          end
+        end
+      end
+
+      private
+
+      def retrieve_artifact_names(name)
+        objs = s3_client.list_objects(bucket: @bucket_name, prefix: name)
+        valid_keys = []
+        objs.contents.each do |o|
```
askreet: FWIW:
```ruby
objs.contents.map(&:key).select { |k| k =~ /^#{name}[\.-]/ }

File: lib/plugins/cleaner.rb [hunk 12]
```diff
@@ -0,0 +1,51 @@
...
+          valid_keys << o.key if o.key =~ /^#{name}[\.-]/
```
askreet: So I'm a bit worried about this logic but I don't have a better suggestion at the moment. Imagine two stacks `dev-chris` and `dev-chris-test1`. Deleting the first would delete all artifacts created for the second so far. Maybe we want `S3Bucket` to start including S3 metadata in the objects?

codwazny: Good catch.

lib/plugins/cleaner.rb
Skarlso: I doubt that a stack would have a 1000 items in archive... no? I searched for the most names with stack names, they were 400 tops or something like that. Especially if this baby is running all the time it won't get that far. What do you think?

codwazny: This is an open source plugin. You can't judge the amount of releases based on our internal use-case. We have no idea how many releases other people will do to their stacks -- it could be one every minute. We have to assume that we will hit that case and it seems very easy to deal with now.

Skarlso: Fair enough. :) But I wasn't only thinking internally. Like... I don't think it's healthy to save a 1000! artifacts for one stack. Like... I don't think that's a good idea in any case. Maybe if it's like, a long running stack and continuously deploy to it, which creates artifacts and you are running for a 100 days now. Okay then. :) 

Anyhow. Sure, I'll deal with it now rather than later. :)

lib/plugins/cleaner.rb
Skarlso: Yeah, that line was to filter out the -213131434.tar.gz stuff and stages which don't have that. So we either look at ways to tag an artifact with a stage name, or we use something like -\d+. :/ Which is ugly.

codwazny: I think the \d method would still have the issue though. `dev-chris` and `dev-chris-1` would still collide. The only way around that is to anchor better to the end but then we have to know the file extensions. Are they always .tar.gz? If so, can we improve it by just making the regex tighter?

Skarlso: They are tar.gz now, but what if we would like to support multiple archiving methods? That would be a b*tch.... :/ we can make it tighter, but we'll loose flexibility.

codwazny: Looking at http://docs.aws.amazon.com/codedeploy/latest/userguide/limits.html it looks like we should accept .tar.gz, .tar, and .zip as those are the only ones allowed for use with code deploy. Unless we want this plugin to be more generic and not tied specifically to cleaning up only code deploy artifacts.

Yitsushi: `\d` is not enough. Use of the above example from @askreet : `dev-chris` and `dev-chris-test1` and `dev-chris-123412134`

How can a filter determine if test1 is a different name or just a version number for `dev-chris`? If I define a fixed version (named)  as `test` for my `dev-yitsushi` stage it will be `dev-yitsushi-test.tar.gz` but if I create a stage as `dev-yitsushi-test` it will be `dev-yitsushi-test-test.tar.gz`

Btw it can be a problem in general. There is 2 stage with a shared S3 bucket:
 - `mysite`
 - `mysite-test`

I create a version for `mysite-test` with a single digit `1` and create a `test-1` version for `mysite` => both will be `mysite-test-1`.

Yitsushi: **[Theoretical level]**

I know it's a much harder way to manage and update the codebase, but what if we can store artifacts with directory stucture?

**Now:**
```
/dev-chris-version1.tar.gz
/dev-chris-version2.tar.gz
/dev-chris-test1-version1.tar.gz
/dev-chris-test1-version2.tar.gz
```

**After:**
```
/dev-chris/dev-chris-version1.tar.gz
/dev-chris/dev-chris-version2.tar.gz
/dev-chris-test1/dev-chris-test-version1.tar.gz
/dev-chris-test1/dev-chris-test-version2.tar.gz
```

With this structure we can easily find `dev-chris` and delete artifacts under that directory without conflict on `dev-chris-test1`

lib/moonshot.rb
```diff
@@ -43,6 +43,8 @@ module Plugins # rubocop:disable Documentation
 ].each { |f| require_relative "moonshot/#{f}" }
 
 # Bundled plugins
+# rubocop:disable Style/WordArray
```
Skarlso: I don't know why this is an offense but the one above is not.

Skarlso: I disabled this check here to not break the style of using an array here. I think this would look odd. :P
```
%w(
  'backup'
  'cleaner'
).each { |p| require_relative "plugins/#{p}" }
```

lib/plugins/cleaner.rb
codwazny: That does sound like a good idea. I wonder why we didn't use a directory to begin with? Perhaps @askreet remembers?

Skarlso:
More tests are incoming.

Skarlso:
Made the regex a whole lot more stricter. 

Now, if you have a stage called ```test``` and it deletes artifacts for it, ```test-test1.tar.gz``` will no longer be found and removed.

lib/plugins/cleaner.rb
askreet: Builds aren't actually related to a stack though, so putting them into a
directory doesn't make sense for moving through environments in a CI
workload. You would create a named build in dev, move through staging and
load testing environments for example then deploy to prod. That all has
nothing to do with the original stack it was built for.

Yitsushi: Make sense, but now 2 stack can create named articafts with the same name (as I described above). I know it's an edge-case but possible.

askreet:
Chatted with @codwazny a bit more about this today. At this point we think this isn't the direction we want moonshot to take for the general case, especially considering we use named builds for our integration testing environment where the largest S3 usage is coming from. I think we'll implement a simple cleanup script for internal use for now.

PR #116, opened 2016-08-15
Use travis client rather than CLI to wait for the job to complete.

jamesiarmes:
This resolves an issue where `travis logs` would continue to run even after the build had completed.

### DIFF
--- a/lib/moonshot/build_mechanism/travis_deploy.rb
+++ b/lib/moonshot/build_mechanism/travis_deploy.rb
@@ -1,4 +1,7 @@
 require 'moonshot/shell'
+require 'travis'
+require 'travis/pro'
+require 'travis/client/auto_login'
 
 module Moonshot::BuildMechanism
   # This simply waits for Travis-CI to finish building a job matching the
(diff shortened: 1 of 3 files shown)


lib/moonshot/build_mechanism/travis_deploy.rb
```diff
@@ -79,9 +85,24 @@ def wait_for_build(version)
     end
 
     def wait_for_job(job_number)
-      cmd = "bundle exec travis logs #{@cli_args} #{job_number}"
-      # This log tailing fails at the end of the file. travis bug.
-      sh_step(cmd, fail: false)
```
glennpratt: Nit: Should this output step information?

glennpratt:
👍

lib/moonshot/build_mechanism/travis_deploy.rb
```diff
@@ -79,9 +85,24 @@ def wait_for_build(version)
...
+      # Find the travis job.
```
pdrakeweb: Could this block be a helper?  At least, through line 91, to enable its reuse in other methods.

pdrakeweb:
Looks good.

lib/moonshot/build_mechanism/travis_deploy.rb
jamesiarmes: I had considered this, but left it here since it's the only place it's currently used. I could move it to a separate method and only call it here. We could move it to an earlier point if we refactor the other methods to use the client rather than the CLI (which I feel we should do long-term, but is out of the scope of this story).

PR #120, opened 2016-08-18
Add tagging to doctor checks.

pdrakeweb:


lib/moonshot/controller.rb
```diff
@@ -67,13 +67,13 @@ def delete
...
       success = true
-      success &&= stack.doctor_hook
-      success &&= run_hook(:build, :doctor)
-      success &&= run_hook(:repo, :doctor)
-      success &&= run_hook(:deploy, :doctor)
+      success &&= stack.doctor_hook(options)
+      success &&= run_hook(:build, :doctor, options)
+      success &&= run_hook(:repo, :doctor, options)
+      success &&= run_hook(:deploy, :doctor, options)
       results = run_plugins(:doctor)
```
askreet: I think we want plugins to support this, too? I am not sure how that would work though.

pdrakeweb: Agreed but this appears to be unused and how to make that happen wasn't readily obvious to me.

PR #121, opened 2016-08-28
Allow the deployment group name to be configurable in order to support and alternate multi-ASG strategy

cpliakas:
**Feature request**

I'd like the deployment group name to be configurable in order to support an admittedly unorthodox usage of Moonshot. This pull request adds that capability.

**Technique**

The unorthodox usage is to manage the CodeDeploy application in CloudFormation, using a single application for the stack and different deployment groups for the ASGs. This allows me to delete the entire stack (including the CodeDeploy stuff) outside of Moonshot if I choose to do so which is sometimes more convenient.

I have a ruby wrapper `bin/environment` that is only used for managing the infrastructure. I then have ruby wrappers for the API and the Worker tier at `bin/api` and `bin/worker` that are used for deploying code for the components. I could be doing this totally wrong, but I really like the workflow and the separation of the infrastructure / app tooling.

### DIFF
--- a/lib/moonshot/deployment_mechanism/code_deploy.rb
+++ b/lib/moonshot/deployment_mechanism/code_deploy.rb
@@ -99,7 +99,7 @@ class Moonshot::DeploymentMechanism::CodeDeploy # rubocop:disable ClassLength
     @app_name || stack.name
   end
 
-  # By default, use the stack name as the geployment group name, unless one has
+  # By default, use the stack name as the deployment group name, unless one has
   # been provided.
   def group_name
     @group_name || stack.name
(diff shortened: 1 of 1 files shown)


askreet:
Hey @cpliakas, sorry I have not seen this PR until now. I need to get much better at paying attention to this repo, my apologies!

This is interesting. I've long known that the concept of your infrastructure having multiple pieces of software deployed to it is something Moonshot is poor at handling. We have at least one team internally that has written a series of tooling to wrap Moonshot to handle their two different tiers of hardware. I have yet to come up with an approach to fixing this problem I think is elegant enough, so here we are.

Looked over the code briefly and it seems to maintain the same default behavior. In my copious spare time, I'd love to chat about your deployment strategy for api+worker (a very common use case), however in the meantime I see no risk in including this feature. I suspect any change to Moonshot to support multiple applications/DGs/builds combining together to produce an environment would be a major version release, anyway.

TL;DR: LGTM, but let me check it out locally and run some use cases against it.

askreet:
All my tests worked locally, so I'm going to merge this. Thanks!

PR #124, opened 2016-09-14
Add documentation concept for ASG Rollout feature.

askreet:
I'm beginning development on a plugin to core moonshot that will enable a command which rolls out launch configuration update in environments where using something as simple as CloudFormation's UpdatePolicy is not an option.

I wrote the documentation first, just to get some feedback and solidify my ideas. I don't intend to commit this PR without an implementation, just soliciting feedback.

Thanks!

### DIFF
--- a/docs/plugins/asg_rollout.md
+++ b/docs/plugins/asg_rollout.md
@@ -1,0 +1,90 @@
+# Auto-Scaling Group LaunchConfig Rollout Tool
+
+## Overview
+
+This plugin adds support for rolling out changes to Auto Scaling
+Groups that have happened after a stack update. It supports various
+pre- and post-actions on the instances (using Moonshot's native SSH
+support).
+
+## Example
+
+The ASGRollout class is intended to be used within a method in your
+project's derived class of the Moonshot::CLI class.
+
+```ruby
+#!/usr/bin/env ruby
+
+require 'moonshot'
+
+class MyService < Moonshot::CLI
+  # .. normal configuration ..
+
+  desc :asg_rollout, 'Update instances in the Auto Scaling Group'
+  def asg_rollout
+    ar = Moonshot::Tools::ASGRollout.new do |config|
+      config.controller = controller
+      config.logical_id = 'APIAutoScalingGroup'
+
+      # Trigger the worker process to initiate a clean shutdown.
+      config.predetach = -> { 0 == exec('systemctl stop my-worker-1') }
+
+      # Wait for that clean shutdown to happen.
+      config.terminate_when = -> { 0 == exec('pgrep -f my-worker-1') }
+    end
+    ar.run!
+  end
+end
+
+MyService.start
+```
+
+## Configuration
+
+The ASGRollout object accepts two required keyword arguments:
+  - **controller**: A Moonshot::Controller. If you are using
+    Moonshot::CLI the `controller` method is most often when you want
+    to use.
+  - **logical_id**: The Logical resource ID from the CloudFormation
+    stack to operate on.
+
+By default, the ASGRollout class will build a list of instances with
+non-conforming LaunchConfiguration, then one at a time perform the
+following actions:
+  1. Increase the *Max* and *Desired* sizes of the Auto Scaling Group by one.
+  2. Wait for the new instance to be *InService*.
+  3. Detach a non-conforming instance.
+  4. Wait for the instance to be *OutOfService* or removed from the
+     Auto Scaling Group.
+  5. If the Auto Scaling Group has an associated Elastic Load
+     Balancer, wait for the instance to be *Todo?* there as well.
+  6. Wait for a new instance to replace it.
+  7. Wait for that new instance to be *InService* in the Auto Scaling Group.
+  8. If the Auto Scaling Group has an associated Elastic Load
+     Balancer, wait for the instance to be *InService* there as well.
+  9. Terminate the detached non-conforming instance.
+  10. If there are other non-conforming instances, go to
+     step 3.
+  11. Restore the *Max* and *Desired* sizes of the Auto Scaling Group
+      to their original values.
+
+A number of optional parameters change the behavior to suit your
+project's needs:
+  - **pre_detach** *(Callable)*: This will be run before step 3
+    above. If it returns `false`, the process will be aborted.
+  - **terminate** *(Callable)*: This will be run to execute step 9. If
+    not specified, the default is to use EC2's Terminate API.
+  - **terminate_when** *(Callable)*: This will be run every 5 seconds
+    prior to step 9. Step 9 will continue as soon as this returns
+    `true`.
+  - **terminate_when_timeout** *(Integer)*: (Default 300s) Number of seconds to
+    wait for the terminate_when lambda to return true. After this
+    timeout, the process will be aborted.
+
+For the lambda / callables above, they are evaluated in a context with
+the following methods available:
+  - **exec** *(TODO)*: Run a command on the instance and return the
+    exit code.
+  - **ec2** *(Aws::EC2::Client)*: A configured Aws::EC2::Client for the region we're
+    operating in.
+  - **instance_id** *(String)*: The EC2 instance ID.


docs/plugins/asg_rollout.md
```diff
@@ -0,0 +1,85 @@
...
+following actions:
+  1. Increase the *Max* and *Desired* sizes of the Auto Scaling Group by one.
+  1. Wait for the new instance to be *InService*.
+  1. Detach a non-conforming instance.
+  1. TODO: Do we need to wait for connection draining or is that
+     handled by the ELB detachment mechiansm?
+  1. Wait the duration of the connecti
+  1. Wait for a new instance to replace it.
+  1. Wait for that new instance to be *InService*.
+  1. Terminate the detached non-conforming instance.
```
glennpratt: What does this mean? In our case it's `ssh shutdown`. Does it need to be a lambda as well?

docs/plugins/asg_rollout.md
```diff
@@ -0,0 +1,85 @@
...
+A number of optional parameters change the behavior to suit your
+project's needs:
+  - **pre_detach**: A lambda or other callable that will be run before
+    step 3 above. If it returns `false`, the process will be aborted.
+  - **terminate_when**: A lambda or other callable that will be run
+    every 5 seconds prior to step 8. Step 8 will continue as soon as
+    this returns `true`.
+  - **terminate_when_timeout**: (Default 300s) Number of seconds to
+    wait for any terminate_when lambda to return true. After this
+    timeout, the process will be aborted.
```
glennpratt: For our use case I don't think we will use any of these hooks (not that that's a problem), we detach, wait for replacement, `ssh shutdown`.

docs/plugins/asg_rollout.md
```diff
@@ -0,0 +1,85 @@
...
+  - **controller**: A Moonshot::Controller. If you are using
+    Moonshot::CLI the `controller` method is most often when you want
+    to use.
+  - **logical_id**: The Logical resource ID from the CloudFormation
+    stack to operate on.
+
+By default, the ASGRollout class will build a list of instances with
+non-conforming LaunchConfiguration, then one at a time perform the
+following actions:
+  1. Increase the *Max* and *Desired* sizes of the Auto Scaling Group by one.
```
gergelyborsothy: `Max` will be **implicitly** increased if `Max == Desired` and you increase `Desired`, same applies for `Min` (but the other way obviously). We must pay attention to restore these values to the original state, not just decrease them by one at the last step.

Eg.: 
- let's say you have `Desired: 1` and `Max: 1`
- you first increase the desired value, you end up with `Desired: 2` and `Max: 2`
- _then_ you increase max: `Desired: 2` and `Max: 3`
- decrease both at the end, in arbitrary order: `Desired: 1` and `Max: 2`

docs/plugins/asg_rollout.md
```diff
@@ -0,0 +1,85 @@
...
+## Overview
+
+This plugin adds support for rolling out changes to Auto Scaling
+Groups that have happened after a stack update. It supports various
+pre- and post-actions on the instances (using Moonshot's native SSH
+support).
+
+## Example
+
+The ASGRollout class is intended to be used within a method in your
```
gergelyborsothy: For this reason, I don't really like the idea considering this class a plugin, and it being in the `Plugins` module. Correct me if I'm wrong, but it won't work like a plugin in it's current sense.

docs/plugins/asg_rollout.md
askreet: I meant EC2 Terminate when I wrote it -- perhaps we need a lambda for "terminate" that defaults to EC2 Terminate, but can be overwritten.

docs/plugins/asg_rollout.md
askreet: Yup, I think for our product we will use these two hooks. With the implementation of the optional hook above I think we'll support your use case, too.

docs/plugins/asg_rollout.md
askreet: This is interesting, I'll keep an eye on this as I'm implementing.

docs/plugins/asg_rollout.md
askreet: Good point, `Moonshot::Tools::ASGRollout`?

gergelyborsothy: Or maybe a `Mechanism`, but `Tools` is fine I guess.

askreet: Yeah, mechanisms are used by the controller. Man, I regret calling them mechanisms.

askreet:
Force-pushed a new update to the docs. @gergelyborsothy, @glennpratt have a few cycles for a re-review?

docs/plugins/asg_rollout.md
```diff
@@ -0,0 +1,92 @@
...
+  - **logical_id**: The Logical resource ID from the CloudFormation
+    stack to operate on.
+
+By default, the ASGRollout class will build a list of instances with
+non-conforming LaunchConfiguration, then one at a time perform the
+following actions:
+  1. Increase the *Max* and *Desired* sizes of the Auto Scaling Group by one.
+  2. Wait for the new instance to be *InService*.
+  3. Detach a non-conforming instance.
+  4. TODO: Do we need to wait for connection draining or is that
```
gergelyborsothy: [AWS](http://docs.aws.amazon.com/elasticloadbalancing/latest/classic/config-conn-drain.html): > While in-flight requests are being served, the load balancer reports the state of a de-registering instance as `InService: Instance deregistration currently in progress` → makes me think it's sort of automatic.

docs/plugins/asg_rollout.md
```diff
@@ -0,0 +1,92 @@
...
+     handled by the ELB detachment mechiansm?
+  5. Wait the duration of the connecti
```
gergelyborsothy: Nit: couple of typos in this line and the one above

docs/plugins/asg_rollout.md
```diff
@@ -0,0 +1,92 @@
...
+project's needs:
+  - **pre_detach** *(Callable)*: This will be run before step 3
+    above. If it returns `false`, the process will be aborted.
+  - **terminate** *(Callable)*: This will be run to execute step 9. If
+    not specified, the default is to use EC2's Terminate API.
+  - **terminate_when** *(Callable)*: This will be run every 5 seconds
+    prior to step 9. Step 9 will continue as soon as this returns
+    `true`.
+  - **terminate_when_timeout** *(Integer)*: (Default 300s) Number of seconds to
+    wait for any terminate_when lambda to return true. After this
```
gergelyborsothy: This means that you can supply multiple lambdas for `terminate_when`?

docs/plugins/asg_rollout.md
```diff
@@ -0,0 +1,92 @@
...
+    timeout, the process will be aborted.
+
+For the lambda / callables above, they are evaluated in a context with
+the following methods available:
+  - **instance** *(TODO)*: An SSH connection to the instance
+    configured using the same logic as Moonshot's SSH feature. (TODO:
+    What class is this? Net::SSH::Thing?)
```
gergelyborsothy: Simple `exec` [here](https://github.com/acquia/moonshot/blob/master/lib/moonshot/stack.rb#L96-L106).

docs/plugins/asg_rollout.md
```diff
@@ -0,0 +1,92 @@
...
+  - **ec2** *(Aws::EC2::Client)*: A configured Aws::EC2::Client for the region we're
+    operating in.
+  - **instance_id** *(String)*: The EC2 instance ID.
```
gergelyborsothy: I guess this is obtainable from the `ec2` resource isn't it? Or just for the convenience's sake?

docs/plugins/asg_rollout.md
askreet: So it stays InService on the ELB, but what about the ASG? Sounds like I want to wait for it to be out of service on the ASG and ELB (if present), much like steps 7 and 8, below.

docs/plugins/asg_rollout.md
askreet: Good catch, that's not my intent. We could support that in the future, though.

docs/plugins/asg_rollout.md
askreet: right this code will have to include something like Net::SSH, or we have to fork/exec to SSH using the same method?

docs/plugins/asg_rollout.md
askreet: There is no EC2 resource listed here, though? We have a client, and an SSH connection. Maybe instead we should just have an exec method that can run a remote command (using fork/exec? for consistency with the existing feature, or something like Net::SSH), and drop the instance method.

askreet: Actually, I like that because it's relatively simple.

askreet:
Latest draft is up @glennpratt and @gergelyborsothy, I'm feeling like I can proceed to implementation at this point but let me know if you have additional feedback.

docs/plugins/asg_rollout.md
```diff
@@ -0,0 +1,92 @@
...
+#!/usr/bin/env ruby
+
+require 'moonshot'
+
+class MyService < Moonshot::CLI
+  # .. normal configuration ..
+
+  desc :asg_rollout, 'Update instances in the Auto Scaling Group'
+  def asg_rollout
+    Moonshot::Tools::ASGRollout.new(
```
askreet: Swapping this for a Config object to line up with other Moonshot internals, and to be sane.

docs/plugins/asg_rollout.md
gergelyborsothy: `Net::SSH` would probably be better here.

docs/plugins/asg_rollout.md
gergelyborsothy: Sorry I thought it's an EC2 resource, not a client. Maybe it'd be more convenient to pass an EC2 resource instead of passing a client and an instance ID. It is possible that passing a client is required to perform operations which are only possible via the client. I'm probably too tired to judge this, so I'll just leave it to you.

gergelyborsothy:
Looks good, I think you can proceed 👍

docs/plugins/asg_rollout.md
```diff
@@ -0,0 +1,90 @@
...
+  - **terminate_when_timeout** *(Integer)*: (Default 300s) Number of seconds to
+    wait for the terminate_when lambda to return true. After this
+    timeout, the process will be aborted.
+
+For the lambda / callables above, they are evaluated in a context with
+the following methods available:
+  - **exec** *(TODO)*: Run a command on the instance and return the
+    exit code.
+  - **ec2** *(Aws::EC2::Client)*: A configured Aws::EC2::Client for the region we're
+    operating in.
```
glennpratt: Nit: If you return `Aws::EC2::Resource`, you will have access to the high level API and the Client.

docs/plugins/asg_rollout.md
```diff
@@ -0,0 +1,90 @@
...
+    prior to step 9. Step 9 will continue as soon as this returns
+    `true`.
+  - **terminate_when_timeout** *(Integer)*: (Default 300s) Number of seconds to
+    wait for the terminate_when lambda to return true. After this
+    timeout, the process will be aborted.
+
+For the lambda / callables above, they are evaluated in a context with
+the following methods available:
+  - **exec** *(TODO)*: Run a command on the instance and return the
+    exit code.
```
glennpratt: You should use `Aq::RemoteSystem` here :trollface:

docs/plugins/asg_rollout.md
```diff
@@ -0,0 +1,90 @@
...
+  3. Detach a non-conforming instance.
+  4. Wait for the instance to be *OutOfService* or removed from the
+     Auto Scaling Group.
+  5. If the Auto Scaling Group has an associated Elastic Load
+     Balancer, wait for the instance to be *Todo?* there as well.
+  6. Wait for a new instance to replace it.
+  7. Wait for that new instance to be *InService* in the Auto Scaling Group.
+  8. If the Auto Scaling Group has an associated Elastic Load
+     Balancer, wait for the instance to be *InService* there as well.
+  9. Terminate the detached non-conforming instance.
```
glennpratt: Does it wait for the instance to actually be terminated. I'm not sure there is any value in doing that or not (perhaps looking for leaks).

askreet: It does not. As i'm implementing, I'm thinking there's no value in having both `terminate` and `terminate_when`. The user can implement steps to terminate, which may include waiting or may include running shutdown blindly. I'm going that route.

gergelyborsothy:
:shipit:

askreet:
Okay, closing this out for now. Soon, I will emerge from the shadows with an implementation...

PR #127, opened 2016-09-18
Refactor SSH functionality.

askreet:
This patch removes all SSH logic from the stack, instead implementing a few new classes to handle the separate roles of building a command, choosing an instance to connect to and actually doing the connection (which lives in the Controller, for now).

This was needed while I was implementing #124, because the command execution and building were tightly coupled to the stack, and I need to be able to use the building logic without the choosing logic, and do so in a fork.

Includes a test update for Travis tests to reduce rspec runtime by 9 seconds.

### DIFF
--- a/lib/moonshot/controller.rb
+++ b/lib/moonshot/controller.rb
@@ -1,3 +1,6 @@
+require_relative 'ssh_target_selector'
+require_relative 'ssh_command_builder'
+
 module Moonshot
   # The Controller coordinates and performs all Moonshot actions.
   class Controller # rubocop:disable ClassLength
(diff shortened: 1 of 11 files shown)


lib/moonshot/controller.rb
```diff
@@ -82,8 +85,13 @@ def doctor
 
     def ssh
       run_plugins(:pre_ssh)
-      stack.ssh
-      run_plugins(:post_ssh)
```
askreet: This was removed because it was never reached, `Stack#ssh` does an exec call, which replaces Ruby with SSH.

askreet: I didn't want to tackle the issue of whether or not we actually want pre-ssh/post-ssh hooks in this patch so I left it functionally equivalent.

PR #129, opened 2016-09-19
On Build failure output the whole error

kaushik:
Truncating the failure output has almost always caused trouble in debugging the issue.


askreet:
Looks good to me.

PR #133, opened 2016-09-26
Add support for YAML templates

borsothy:


lib/moonshot/json_stack_template.rb
```diff
@@ -0,0 +1,26 @@
...
+        Parameter.new(k, v['Default'])
+      end
+    end
+
+    def resource_names
+      @body.fetch('Resources', {}).keys
+    end
+
+    def body
+      @body.to_json
```
gergelyborsothy: Introduced this to reduce unneccessary parsing/dumping to string/JSON (or YAML) and vice versa.

lib/moonshot/stack.rb
```diff
@@ -154,12 +155,14 @@ def default_values
...
-      @template ||= StackTemplate.new(template_file, log: @log)
+      @template ||= load_template_file
     end
 
     # @return [String] the path to the template file.
     def template_file
-      File.join(Dir.pwd, 'cloud_formation', "#{@app_name}.json")
+      json = File.join(Dir.pwd, 'cloud_formation', "#{@app_name}.json")
+      yaml = File.join(Dir.pwd, 'cloud_formation', "#{@app_name}.y*ml")
+      @template_file ||= Dir[json].first || Dir[yaml].first || raise
```
gergelyborsothy: In order to maintain compatibility, I had to introduce this. The problem with this approach is the `load_template_file` function. `json_template.exists?` will return true if the template file is actually YAML formatted, but will die on parsing. We probably need to rethink the way to select the right object to construct.

gergelyborsothy: I also wanna raise a custom exception here, what do you think?

lib/moonshot/stack_template.rb
```diff
@@ -10,30 +8,19 @@ def required?
...
-      # formatting white space.
-      @body = JSON.parse(File.read(filename)).to_json
+      @filename = Dir[filename].first
     end
 
-    def parameters
-      JSON.parse(@body).fetch('Parameters', {}).map do |k, v|
-        Parameter.new(k, v['Default'])
-      end
+    def validate
```
gergelyborsothy: I introduced this function, I don't know if we need it or not (probably out of scope of this PR).

lib/moonshot/stack.rb
gergelyborsothy: Maybe we could flip the logic here, and store the path in the template object itself.

gergelyborsothy:
Also, I guess we need more tests.

lib/moonshot/stack_template.rb
```diff
@@ -1,7 +1,5 @@
-require 'json'
-
 module Moonshot
-  # A StackTemplate loads the JSON template from disk and stores information
+  # A StackTemplate loads the template from disk and stores information
   # about it.
   class StackTemplate
```
gergelyborsothy: I _really_ wanna make this class abstract.

lib/moonshot/stack.rb
```diff
@@ -195,6 +198,19 @@ def load_parameters_file
       end
     end
 
+    def load_template_file
+      json_template = JsonStackTemplate.new(template_file, log: @log)
```
askreet: Rather than depending on a JSON parser error in the YAML file, I think you want to do something like:
```ruby
json_template = JsonStackTemplate.new(json_template_file, log: @log)
yaml_template = YamlStackTemplate.new(yaml_template_file, log: @log)
```

In this way, the `#exists?` method can decide if the class found a file or not. This would, of course, mean lazy-loading the file only when necessary, rather than in the constructor, which might be ideal anyway.

lib/moonshot/yaml_stack_template.rb
```diff
@@ -0,0 +1,26 @@
...
+    end
+
+    def parameters
+      @body['Parameters'].map do |k, v|
+        Parameter.new(k, v['Default'])
+      end
+    end
+
+    def resource_names
+      @body['Resources'].keys
```
askreet: This is the same as the JSON implementation, except that it fails if there is no "Resources" key at the top level. Is that by design? If not, can we push this up to the parent class with the implementation from your JSON class?

lib/moonshot/yaml_stack_template.rb
```diff
@@ -0,0 +1,26 @@
...
+module Moonshot
+  # Handles YAML formatted AWS template files.
+  class YamlStackTemplate < StackTemplate
+    def initialize(filename, log:)
+      super(filename, log: log)
+      @body = YAML.load_file(@filename) if exists?
+    end
+
+    def parameters
+      @body['Parameters'].map do |k, v|
```
askreet: Same comment as the resource_names method below -- which leads me to believe you have a reason for not using `#fetch` in this class?

lib/moonshot/json_stack_template.rb
```diff
@@ -0,0 +1,26 @@
+require 'json'
+require_relative 'stack_template'
+
+module Moonshot
```
askreet: So, after reviewing all the code in these classes, I suggest:
- Lazy-loading private methods for accessing the template, like:
```ruby
class JsonStackTemplate < StackTemplate
  def body
    template_body.to_json
  end

  private

  def template_body
    @template_body ||= JSON.parse(File.read(@filename))
  end
end
```

- Moving parameters/resource_names into the parent class with the `#fetch` implementations here, since they seem compatible, unless I am missing something.

lib/moonshot/stack_template.rb
```diff
@@ -10,30 +8,19 @@ def required?
...
-        Parameter.new(k, v['Default'])
-      end
+    def validate
+      @body.bytes.length <= MAX_TEMPLATE_BODY_LENGTH_IN_BYTES
     end
 
-    # Return a list of defined resource names in the template.
-    def resource_names
-      JSON.parse(@body).fetch('Resources', {}).keys
+    def exists?
```
askreet: I think we should use `#exist?` to align with the method names in stdlib, Ruby has been deprecating `#exists?`.

lib/moonshot/stack.rb
```diff
@@ -160,8 +160,8 @@ def template
 
     # @return [String] the path to the template file.
     def template_file
-      json = File.join(Dir.pwd, 'cloud_formation', "#{@app_name}.json")
-      yaml = File.join(Dir.pwd, 'cloud_formation', "#{@app_name}.y*ml")
+      json = "#{raw_template_file_name}.json"
+      yaml = "#{raw_template_file_name}.y*ml"
       @template_file ||= Dir[json].first || Dir[yaml].first || raise
```
gergelyborsothy: What exception shall we raise here? I wanna avoid throwing a simple string tbh.

lib/moonshot/stack.rb
```diff
@@ -198,13 +198,18 @@ def load_parameters_file
...
+      yaml_template = YamlStackTemplate.new("#{raw_template_file_name}.y*ml", log: @log)
       case
-      when json_template.exists?
+      when json_template.exist?
         json_template
-      when yaml_template.exists?
+      when yaml_template.exist?
         yaml_template
       else
         raise
```
gergelyborsothy: Same question as above.

gergelyborsothy:
I see there are no separate tests for the original `StackTemplate` class either. Shall I start writing ones for the two new subclasses regardless?

lib/moonshot/stack.rb
```diff
@@ -195,6 +198,24 @@ def load_parameters_file
...
     end
 
+    # @return [String] the path to the parameters file without extension.
+    def raw_template_file_name
+      @raw_template_file_name ||= File.join(Dir.pwd, 'cloud_formation', @app_name)
+    end
+
+    def load_template_file
+      json_template = JsonStackTemplate.new("#{raw_template_file_name}.json", log: @log)
+      yaml_template = YamlStackTemplate.new("#{raw_template_file_name}.y*ml", log: @log)
```
askreet: We already dictate that parameters files end in .yml, without support for .yaml -- I'd prefer to just do the same here. It simplifies documentation and code.

lib/moonshot/stack.rb
```diff
@@ -154,12 +155,14 @@ def default_values
...
-      @template ||= StackTemplate.new(template_file, log: @log)
+      @template ||= load_template_file
     end
 
     # @return [String] the path to the template file.
     def template_file
-      File.join(Dir.pwd, 'cloud_formation', "#{@app_name}.json")
+      json = "#{raw_template_file_name}.json"
+      yaml = "#{raw_template_file_name}.y*ml"
+      @template_file ||= Dir[json].first || Dir[yaml].first || raise
```
askreet: Is this still used anywhere after these changes?

askreet: If so, I think you could use `load_template_file.filename` here, if you make an attr_reader on `@filename`.

lib/moonshot/stack.rb
```diff
@@ -195,6 +198,24 @@ def load_parameters_file
...
+    def load_template_file
+      json_template = JsonStackTemplate.new("#{raw_template_file_name}.json", log: @log)
+      yaml_template = YamlStackTemplate.new("#{raw_template_file_name}.y*ml", log: @log)
+      case
+      when json_template.exist?
+        json_template
+      when yaml_template.exist?
+        yaml_template
+      else
+        raise
```
askreet: RE: your question about exceptions. I generally avoid making new exception classes unless I actually expect someone would use them to implement recovery logic. In this case, `RuntimeError` seems appropriate, because we're just going to rewrite this to a Thor::Error in the CLI class.

lib/moonshot/yaml_stack_template.rb
```diff
@@ -0,0 +1,21 @@
+require 'yaml'
+require_relative 'stack_template'
+
+module Moonshot
+  # Handles YAML formatted AWS template files.
+  class YamlStackTemplate < StackTemplate
+    def initialize(filename, log:)
+      super(filename, log: log)
```
askreet: Can drop this method now, and just use the inherited method.

lib/moonshot/json_stack_template.rb
```diff
@@ -0,0 +1,21 @@
...
+  # Handles JSON formatted AWS template files.
+  class JsonStackTemplate < StackTemplate
+    def initialize(filename, log:)
+      super(filename, log: log)
```
askreet: Same here.

askreet:
@borsothy re: test coverage -- We have some tests which already use FakeFS so it should be trivial to cover the fallback strategy and exception behaviors when there is no template. I'd love to see some tests, personally.

lib/moonshot/stack.rb
borsothy: In plugins?

lib/moonshot/stack.rb
borsothy: Okay.

lib/moonshot/stack.rb
askreet: We can cut 0.8.0 to signify the breaking change. However, if we're going to do that I wonder if we want to even support this type of access. Instead, we'd probably want to act as an abstraction layer for the plugins, so they don't have to do File I/O and serialize/deserialize. Plus we're caching the file now, meaning if they do change the file later actions won't be affected. This would definitely break any plugins using that kind of modification of the template.

I don't see that we bundle any plugins that use this feature, nor is it used in the one internal plugin I know of. I feel comfortable saying we'd rather provide interfaces to specific data (like `#resources` and `#parameters` in this class), than give plugins the file to muck with directly, anyway.

borsothy:
@askreet let me know if you're okay with the code, and I'll squash the commits.

askreet:
Reviewing stuff is really noisy with the new github view not clearing out old comments. I'm going to delete outdated comments and re-review. Just a heads up.

lib/moonshot/stack.rb
```diff
@@ -154,12 +155,15 @@ def default_values
...
+      @template ||= load_template_file
     end
 
     # @return [String] the path to the template file.
     def template_file
-      File.join(Dir.pwd, 'cloud_formation', "#{@app_name}.json")
+      json = "#{raw_template_file_name}.json"
+      yaml = "#{raw_template_file_name}.yml"
+      @template_file ||= Dir[json].first || Dir[yaml].first || \
+                         raise('No conforming template file was found.')
```
askreet: I like to always give the user full context so they're not confused about what is expected. What about:
```ruby
@template_file ||= Dir[json].first || Dir[yaml].first
raise "CloudFormation template not found at #{json} or #{yaml}!" unless @template_file
```

lib/moonshot/stack.rb
```diff
@@ -195,6 +199,24 @@ def load_parameters_file
       end
     end
 
+    # @return [String] the path to the parameters file without extension.
```
askreet: nit: This isn't the parameters file, it's the template file.

lib/moonshot/stack.rb
```diff
@@ -195,6 +199,24 @@ def load_parameters_file
...
+    def load_template_file
+      json_template = JsonStackTemplate.new("#{raw_template_file_name}.json", log: @log)
+      yaml_template = YamlStackTemplate.new("#{raw_template_file_name}.yml", log: @log)
+      case
+      when json_template.exist?
+        json_template
+      when yaml_template.exist?
+        yaml_template
+      else
+        raise 'No conforming template file was found.'
```
askreet: Use the same exception here as I described above.

spec/moonshot/stack_spec.rb
```diff
@@ -139,4 +139,35 @@
       expect(subject.parameters_file).to eq(path)
     end
   end
+
+  describe '#template' do
```
askreet: Nice test 👍 !

borsothy: http://i2.kym-cdn.com/entries/icons/facebook/000/006/077/so_good.jpg

lib/moonshot/stack_template.rb
glennpratt: Make the base initialize raise and move whatever common work it does into a helper.

[NotImplementedError](https://ruby-doc.org/core-2.2.0/NotImplementedError.html) sounds kinda right here, but of course isn't. There's probably something appropriate-ish.

borsothy:
@askreet squashed ^

askreet:
This is looking good to me -- going to pull it down and do some tests against a real-world project.

askreet:
This worked just fine. As an aside, I loved how easy it was to convert from JSON to YAML...

Issue #131, opened 2016-09-23
RFC: Naming of build-version, deploy-version and deploy-code.


askreet:
I've been using Moonshot extensively for our project in the last 3 months and I've found that the naming of these commands is overly verbose, and the conflict between the short forms of `deploy-version` and `deploy-code` is annoying.

I'm looking for feedback on renaming, or aliasing, these commands to:
```
build-version -> build
deploy-version -> deploy
deploy-code -> dev-deploy, devdeploy, localdeploy, ???
```

I can't think of a perfect answer for `deploy-code`, but when managing environments across out service delivery pipeline, I definitely wish the first two were as describe above.

pdrakeweb:
I would suggest renaming deploy-code to deploy and leaving deploy-version and build-version alone.  These could have short aliases via map (d, v and b respectively?).  deploy-code could be left as an alias for backward compatibility.

askreet:
@pdrakeweb I don't love the one-letter options, I feel they're really easy to typo. Honestly, as pedantic as it sounds, I'd feel a bit better with two-letter options.

That said, I was trying to make the commands used during the movement of a version through post-development environments shorter, not `deploy-code`.

What about this?
```
build-version -> build
deploy-version -> deploy
deploy-code -> push
```

In all cases, leave the previous full names as aliases, no backwards compatibility is broken.

pdrakeweb:
I like one-letter command aliases but don't feel strongly about them.  95% of the time when I actually *type* a moonshot command it is for development purposes (non-versioned) and so it is most important to me that those be short.

Is there a good reason to change these in Moonshot vs folks simply mapping things how they like in their own CLI tool classes?  Of course, these short command names don't preclude that.

askreet:
@pdrakeweb I think a good reason is simplifying mobility between teams using the same SDLC tooling. It's also a reason I want to revive #89.

askreet:
This was addressed in #139

PR #139, opened 2016-10-04
Moonshot 1.0.0 - multiple features.

askreet:
This is the starting point for a number of discussions and issues. I'm starting to wonder how much value we're getting out of Thor and if stdlib's OptionParser is a better use of our time.

Looking for some early feedback on these changes.


lib/moonshot/command_line.rb
```diff
@@ -0,0 +1,104 @@
...
+
+      # Include all Moonshot::Command and Moonshot::SSHCommand
+      # derived classes as subcommands, with the description of their
+      # default task.
+      registered_commands.each do |klass|
+        default_task = klass.default_task
+        description = klass.tasks.fetch(default_task).description rescue ''
+        command_name = klass.to_s.split('::').last.scan(/[A-Z][a-z]+/).map(&:downcase).join('-')
+
+        desc command_name, description
```
askreet: I cannot get this to work with commands that are hyphen separated with subclasses. Also unhappy with how I lazy-load the description from the default task after the fact. Thinking that if I drop Thor I can make this do what I want explicitly with some custom code and OptionParser.

lib/moonshot/controller_config.rb
```diff
@@ -1,5 +1,6 @@
 require_relative 'default_strategy'
 require_relative 'ssh_config'
+require_relative 'task'
```
askreet: This whole "task" interface was from an earlier attempt and will be removed.

askreet:
This is focusing on #131 and #89.

lib/moonshot/command_line.rb
glennpratt: I'd like to play with this a bit.

glennpratt: (No idea when that will be though :smile: )

askreet: Well, you can race me to replacing this with OptionParser. Whomever gets there first wins. :-)

lib/moonshot/command_line.rb
```diff
@@ -0,0 +1,108 @@
...
+        end
+      end
+
+      moonfile_dir = Dir.pwd
+      Dir.chdir(orig_dir)
+
+      # Load any plugins and CLI extensions relative to the Moonfile
+      if File.directory?(File.join(moonfile_dir, 'moonshot'))
+        load_cli_extensions(moonfile_dir)
+        load_plugins(moonfile_dir)
```
glennpratt: How can shared extensions be registered?

lib/moonshot/command_line.rb
```diff
@@ -0,0 +1,108 @@
...
+    end
+
+    def self.usage
+      warn "Usage: moonshot [command]"
+      warn
+      warn "Valid commands include:"
+      @commands.each do |c, k|
+        warn "#{c} #{k.description}"
+      end
+      exit 1
```
glennpratt: I would separate usage from the exit 1 (invalid arguments).

`cmd help` or `--help` is the first thing I'm going to do and that's not an error.

lib/moonshot/command_line.rb
```diff
@@ -0,0 +1,108 @@
...
+      end
+    end
+
+    def self.usage
+      warn "Usage: moonshot [command]"
+      warn
+      warn "Valid commands include:"
+      @commands.each do |c, k|
+        warn "#{c} #{k.description}"
+      end
```
glennpratt: This looked a little dense compared to other tools, some whitespace might be nice.

Global options would be nice too.

```
$ cargo --help
Rust's package manager

Usage:
    cargo <command> [<args>...]
    cargo [options]

Options:
    -h, --help          Display this message
    -V, --version       Print version info and exit
    --list              List installed commands
    --explain CODE      Run `rustc --explain CODE`
    -v, --verbose ...   Use verbose output
    -q, --quiet         No output printed to stdout
    --color WHEN        Coloring: auto, always, never

Some common cargo commands are (see all commands with --list):
    build       Compile the current project
    clean       Remove the target directory
    doc         Build this project's and its dependencies' documentation
    new         Create a new cargo project
    init        Create a new cargo project in an existing directory
    run         Build and execute src/main.rs
    test        Run the tests
    bench       Run the benchmarks
    update      Update dependencies listed in Cargo.lock
    search      Search registry for crates
    publish     Package and upload this project to the registry
    install     Install a Rust binary

See 'cargo help <command>' for more information on a specific command.

$ composer --help
Usage:
 help [--xml] [--format="..."] [--raw] [command_name]

Arguments:
 command               The command to execute
 command_name          The command name (default: "help")

Options:
 --xml                 To output help as XML
 --format              To output help in other formats (default: "txt")
 --raw                 To output raw command help
 --help (-h)           Display this help message
 --quiet (-q)          Do not output any message
 --verbose (-v|vv|vvv) Increase the verbosity of messages: 1 for normal output, 2 for more verbose output and 3 for debug
 --version (-V)        Display this application version
 --ansi                Force ANSI output
 --no-ansi             Disable ANSI output
 --no-interaction (-n) Do not ask any interactive question
 --profile             Display timing and memory usage information
 --working-dir (-d)    If specified, use the given directory as working directory.

Help:
 The help command displays help for a given command:
   composer help list

 You can also output the help in other formats by using the --format option:

   composer help --format=xml list

 To display the list of available commands, please use the list command.
```

askreet:
I've got to forward port some new features into this patch. The original branch is very outdated.

lib/moonshot/command.rb
```diff
@@ -0,0 +1,67 @@
...
+        o.on('-nNAME', '--environment=NAME', 'Which environment to operate on.') do |v|
+          Moonshot.config.environment_name = v
+        end
+
+        o.on('--[no-]interactive-logger', TrueClass, 'Enable or disable fancy logging') do |v|
+          @use_interactive_logger = v
+        end
+
+        # TODO: This is going to be shown for incompatible options, we
+        # could create more class hierarchy to support it?
```
glennpratt: A class hierarchy is probably fine, could also be a method (add_stack_options(parser)) or module if you want to avoid hierarchy.

I would suggest the ones at the top level be considered global and can be printed in the top level help.

lib/moonshot/command_line.rb
```diff
@@ -0,0 +1,108 @@
...
+    def self.load_commands
+      @commands = {}
+
+      # Include all Moonshot::Command and Moonshot::SSHCommand
+      # derived classes as subcommands, with the description of their
+      # default task.
+      registered_commands.each do |klass|
+        next unless klass.instance_methods.include?(:execute)
+
+        command_name = klass.to_s.split('::').last.scan(/[A-Z][a-z]+/).map(&:downcase).join('-')
```
glennpratt: I'm not entirely sure what's this is supposed to do.

```ruby
[1] pry(main)> base = "My::Project::"
=> "My::Project::"
[2] pry(main)> class_name = "#{base}::Foo"
=> "My::Project::::Foo"
[3] pry(main)> class_name.split('::').last.scan(/[A-Z][a-z]+/).map(&:downcase).join('-')
=> "foo"
[4] pry(main)> class_name = "#{base}::S3Browser"
=> "My::Project::::S3Browser"
[5] pry(main)> class_name.split('::').last.scan(/[A-Z][a-z]+/).map(&:downcase).join('-')
=> "browser"
```

lib/moonshot/command_line.rb
askreet: Do you mean like one that is included via a gem? Requiring the gem, which would subclass Command/SSHCommand would load them in. I've got a branch for one of the internal shared extensions that I'm using locally, I'll send you a link.

lib/moonshot/command_line.rb
askreet: Definitely, I plan to add that, thanks.

lib/moonshot/command_line.rb
glennpratt: Got it, this just loads the ruby code, resolved.

lib/moonshot/command_line.rb
askreet: Hm, the regexp makes the assumption that your UpperCamelCaseWords aren't a capital followed by a number.

Ideally:
S3Bucket -> s3-bucket
SSHThingDoer -> ssh-thing-doer
MyFavoriteThing -> my-favorite-thing

Need to refine this a bit.

lib/moonshot/command_line.rb
```diff
@@ -0,0 +1,108 @@
...
+      plugins_path = File.join(moonfile_dir, 'moonshot', 'plugins', '**', '*.rb')
+      Dir.glob(plugins_path).each do |file|
+        load(file)
+      end
+    end
+
+    def self.load_cli_extensions(moonfile_dir)
+      cli_extensions_path = File.join(moonfile_dir, 'moonshot', 'cli_extensions', '**', '*.rb')
+      Dir.glob(cli_extensions_path).each do |file|
+        load(file)
```
glennpratt: Since these are `.rb` files, could you `require_relative` them? Should help prevent double loads.

lib/moonshot/command_line.rb
```diff
@@ -0,0 +1,108 @@
...
+      # Load any plugins and CLI extensions relative to the Moonfile
+      if File.directory?(File.join(moonfile_dir, 'moonshot'))
+        load_cli_extensions(moonfile_dir)
+        load_plugins(moonfile_dir)
+      end
+
+      Object.include(Moonshot::ArtifactRepository)
+      Object.include(Moonshot::BuildMechanism)
+      Object.include(Moonshot::DeploymentMechanism)
+      load(File.join(moonfile_dir, 'Moonfile'))
```
askreet: We could support multiple names, like Rake does -- but I wonder if we should have a preference. I tend to agree that if this is ruby code it ends in .rb.

lib/moonshot/command_line.rb
glennpratt: Just a pinch of active support :)

https://github.com/acquia/cloud-data-api-client-ruby/blob/98b58b60134b24a152e3a1fe8130185fe3a1b20b/lib/acquia/cloud/database/tools/base.rb#L40-L47

lib/moonshot/command_line.rb
glennpratt:
```
git mv Rakefile Rakefile.rb
rake -T
rake ah-config:spec                   # Run ah-config specs
rake ah-tasks                         # Run default `ah-tasks` development tasks
...
```
Nice

IMO, just do .rb. You will probably never need to defend that decision.

lib/moonshot/command_line.rb
glennpratt: I point out active support because the common case everyone gets wrong is "Ruby-cased" acronyms.

```
=> "My::Project::::HTTPClient"
[4] pry(main)> class_name.split('::').last.scan(/[A-Z][a-z]+/).map(&:downcase).join('-')
=> "client"
```

glennpratt:
Something that might be nice is to convert the args when argv[1] is help

```
moonshot help ssh -> ssh --help
```

Or just error with the extra arg and disply the supported order in help

lib/moonshot/command_line.rb
askreet: Addressed in last push, now raises a RuntimeError and there is a handler in bin/moonshot.

lib/moonshot/command_line.rb
askreet: Addressed in last push.

lib/moonshot/command_line.rb
askreet: Addressed in the last patch. `S3BucketSSLUploader` now works as expected. :-)

lib/moonshot/command_line.rb
```diff
@@ -0,0 +1,121 @@
...
+
+      # Load any plugins and CLI extensions relative to the Moonfile
+      if File.directory?(File.join(moonfile_dir, 'moonshot'))
+        load_cli_extensions(moonfile_dir)
+        load_plugins(moonfile_dir)
+      end
+
+      Object.include(Moonshot::ArtifactRepository)
+      Object.include(Moonshot::BuildMechanism)
+      Object.include(Moonshot::DeploymentMechanism)
```
glennpratt: I never noticed this before. Is this to shorten the constant lookup in the Moonfile?

If so, do you only need that to happen inside the config block? Just wondering about a solution that wouldn't touch Object, but it's not critical.

askreet: I'd be happy to address this in a follow-up.

PR #150, opened 2016-10-12
Introduced 'new' command.

borsothy:


lib/moonshot/command_line.rb
```diff
@@ -4,12 +4,23 @@ module Moonshot
   # This class implements the command-line `moonshot` tool.
   class CommandLine
     def self.run! # rubocop:disable AbcSize, CyclomaticComplexity, MethodLength, PerceivedComplexity
+      # Handling premature early command like
```
askreet: I'd like to see a bit more meat here about why this is necessary, like:
```ruby
# Commands defined as Moonshot::Commands require a properly
# configured Moonshot.rb and supporting files. Without them, we only
# support `--help` and `new`.
```

lib/moonshot/command_line.rb
```diff
@@ -4,12 +4,23 @@ module Moonshot
...
+      # `moonshot new` or `moonshot --help`
+      catch(:regular_command) do
```
askreet: I don't have a really compelling reason *why*, but I don't like the use of catch/throw between methods here. Might be simpler if `handle_early_commands` returns true when it has handled an early command and you can simply write:
```ruby
return if handle_early_commands(ARGV[0..1])
```

lib/moonshot/commands/new.rb
```diff
@@ -0,0 +1,82 @@
...
+
+        def generate_moonfile
+          <<-EOF
+Moonshot.config do |m|
+  m.app_name = '#{@application_name}'
+  m.artifact_repository = S3Bucket.new('<your_bucket>')
+  m.build_mechanism = Script.new('bin/build.sh')
+  m.deployment_mechanism = CodeDeploy.new(asg: 'AutoScalingGroup')
+end
+        	EOF
```
askreet: Odd spacing here.

lib/moonshot/commands/new.rb
```diff
@@ -0,0 +1,82 @@
...
+        def fill_moonfile
+          File.open(moonfile_path, 'w') { |f| f.write generate_moonfile }
+        end
+
+        def generate_moonfile
+          <<-EOF
+Moonshot.config do |m|
+  m.app_name = '#{@application_name}'
+  m.artifact_repository = S3Bucket.new('<your_bucket>')
+  m.build_mechanism = Script.new('bin/build.sh')
```
askreet: I wonder if we want to put some language into this file (I can't comment on empty files in GitHub, so I'm commenting here).

Perhaps something like:
```shell
#!/bin/bash
#

echo "Building version '$VERSION_NAME', using target file '$OUTPUT_FILE'."

# Add code here to assemble your project into a compressed tarball (.tar.gz)
# located at $OUTPUT_FILE. Be sure to include your appspec.yml in the tarball's
# root!
#
# For example:
#
# mkdir build_dir
# rsync -avP path-to-source build_dir/src
# cp moonshot/appspec.yml build_dir/
# tar czf $OUTPUT_FILE -C build_dir .
# rm -rf build_dir
#

File: lib/moonshot/commands/new.rb [hunk 5]
```diff
@@ -0,0 +1,82 @@
...
+
+      class << self
+        def run!(application_name)
+          @application_name = application_name
+
+          create_project_dir
+          copy_defaults
+          create_file(parameter_path)
+          create_file(template_path)
+          fill_moonfile
```
askreet: Can we add some output here that says something like:

---
Your application is configured, the following changes have been made to your project directory:
- Created a Moonfile.rb where you can configure your project.
- Created moonshot/plugins where you can add hooks to core Moonshot actions.
- Created moonshot/cli_extensions where you can create project-specific commands.

You will also need to ensure your amazon account is configured for CodeDeploy, by creating a role that allows deployments. See: [link to moonshot documentation].

lib/moonshot/commands/new.rb
```diff
@@ -0,0 +1,82 @@
...
+          Dir.pwd
+        end
+
+        def create_project_dir
+          raise "Directory '#{@application_name}' already exists!" \
+            if Dir.exist?(project_path)
+          Dir.mkdir(project_path)
+        end
+
+        def project_path
```
askreet: I'm torn on if we want `new` to be run inside an existing project directory, or if it should create a project directory. I would expect many users would be importing an existing project as it's getting ready for being deployed on CloudFormation-based stacks, but perhaps that would change over time?

lib/moonshot/command_line.rb
```diff
@@ -4,12 +4,23 @@ module Moonshot
...
+
       # Find the Moonfile in this project.
       orig_dir = Dir.pwd
 
       loop do
         break if File.exist?('Moonfile.rb')
-        raise 'Could not find Moonfile.rb!' if Dir.pwd == '/'
+
+        if Dir.pwd == '/'
+          raise 'No Moonfile.rb found, are you in a project? Maybe you need to '\
```
askreet: I would warn the longer text out, and raise a simpler error. The output we do in the binstub takes the first line of the backtrace so I just think this would be hard to read.

askreet:
@borsothy I added some parts of this ("handle_early_command") to #159, so you will have to rebase. I also made CommandLine into a real class instead of static methods only, because I think it will help us break up that mega-method I created in the future.

lib/moonshot/commands/new.rb
borsothy: Yeah so if I add the correct indentation, it will also appear in the fail. Aware of any way to prevent that?

lib/moonshot/commands/new.rb
borsothy: I would stick with a new directory, that's how most of these commands work (like with rails).

borsothy:
@askreet most changes addressed, couple of questions/answers above. Will squash before merge.

PR #151, opened 2016-10-14
Implement first pass at AWS-as-source-of-truth parameters management.

askreet:
This starts to implement #184, minus the interactive gathering of parameters. Includes testing and I've done quite a bit of testing on my 1.x branch for our project.

### DIFF
--- a/lib/moonshot/commands/create.rb
+++ b/lib/moonshot/commands/create.rb
@@ -1,8 +1,14 @@
+require_relative 'parameter_arguments'
+
 module Moonshot
   module Commands
     class Create < Moonshot::Command
+      include ParameterArguments
+
       self.usage = 'create [options]'
       self.description = 'Create a new environment'
+
+      attr_reader :version, :deploy
 
       def parser
         @deploy = true
--- a/lib/moonshot/controller.rb
+++ b/lib/moonshot/controller.rb
@@ -1,9 +1,13 @@
 require_relative 'ssh_target_selector'
 require_relative 'ssh_command_builder'
 
+require_relative 'stack_parameter'
+require_relative 'parameter_collection'
+require_relative 'parent_stack_parameter_loader'
+
 module Moonshot
   # The Controller coordinates and performs all Moonshot actions.
-  class Controller
+  class Controller # rubocop:disable ClassLength
     attr_accessor :config
 
     def initialize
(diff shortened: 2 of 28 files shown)


askreet:
Ping @borsothy, maybe you have some time to review this? It's actually quite slick in my local testing, much easier to reason about than parameters file. Combined with ChangeSet support for UpdateStack this will create a really nice user experience I think.

borsothy:
Whoa, sure. Give me some time tho 😇

lib/moonshot/controller.rb
```diff
@@ -41,7 +118,8 @@ def status
     end
 
     def deploy_code
-      version = "#{stack_name}-#{Time.now.to_i}"
+      version = [@config.app_name, @config.environment_name, Time.now.to_i]
+                .join('-')
```
borsothy: Is there any length restriction for this? Either for S3 or CD?

lib/moonshot/resources.rb
```diff
@@ -2,11 +2,12 @@ module Moonshot
   # Resources is a dependency container that holds references to instances
   # provided to a Mechanism (build, deploy, etc.).
   class Resources
-    attr_reader :stack, :ilog
+    attr_reader :stack, :ilog, :controller
```
borsothy: 🎉

lib/moonshot/stack_parameter.rb
```diff
@@ -0,0 +1,64 @@
...
+      @value = nil
+    end
+
+    # Does this Stack Parameter have a default value that will be used?
+    def default?
+      !@default.nil?
+    end
+
+    def use_previous?
+      @use_previous ? true : false
```
borsothy: I don't see the scenario where this can be `nil`. Even if that could occur, you can still use `!!@use_previous`. So I wonder why the ternary 🤔

lib/moonshot/stack_parameter.rb
```diff
@@ -0,0 +1,64 @@
...
+    def set?
+      !@value.nil?
+    end
+
+    def set(value)
+      @value = value
+      @use_previous = false
+    end
+
+    def use_previous
```
borsothy: I'd rename this `use_previous!`, it looks like an accessor as it is now.

lib/moonshot/stack_parameter.rb
```diff
@@ -0,0 +1,64 @@
...
+      !@default.nil?
+    end
+
+    def use_previous?
+      @use_previous ? true : false
+    end
+
+    # Has the user provided a value for this parameter?
+    def set?
+      !@value.nil?
```
borsothy: This makes me think if an empty string is considered okay as a value for a parameter.

lib/moonshot/commands/create.rb
```diff
@@ -15,6 +15,19 @@ def parser
...
         end
+
+        parser.on('--answer-file FILE', '-aFILE', 'Load Stack Parameters from a YAML file') do |v|
+          Moonshot.config.answer_file = File.expand_path(v)
+        end
+
+        parser.on('--parameter KEY=VALUE', '-PKEY=VALUE', 'Specify Stack Parameter on the command line') do |v| # rubocop:disable LineLength
+          data = v.split('=')
+          unless data.size == 2
+            raise "Invalid parameter format '#{v}', expected KEY=VALUE (e.g. MyStackParameter=12)"
```
askreet: I think the logic will fail -- this warrants a fix and test, will add. Nice catch!

lib/moonshot/controller.rb
askreet: Perhaps, out of scope of this story. This just keeps compatibility with the current naming.

lib/moonshot/stack_parameter.rb
askreet: Because Rubocop yelled at me for double-negate, and I didn't feel like arguing with it right now :(.

lib/moonshot/stack_parameter.rb
askreet: I like that, will fix. Thanks.

lib/moonshot/stack_parameter.rb
askreet: Hmm, I'm not sure if that is allowed in CF or how to handle it here, let me think about it.

askreet:
Thanks @borsothy, I'll get a new patch up soon.

lib/moonshot/commands/create.rb
askreet: Latest push fixes this and adds some spec coverage around parameters parsing for create/update, as well as putting some duplicate code into a module.

lib/moonshot/controller_config.rb
```diff
@@ -8,23 +8,28 @@ class ControllerConfig
...
     attr_accessor :project_root
+    attr_accessor :parameters
+    attr_accessor :parameter_overrides
 
     def initialize
+      @interactive           = true
       @interactive_logger    = InteractiveLogger.new
-      @parameter_strategy    = Moonshot::ParameterStrategy::DefaultStrategy.new
+      @parameters            = {}
+      @parameter_overrides   = {}
```
askreet: Addressed in latest patch by using a real factory method and sane state in the class.

askreet:
Feedback addressed -- merging this. Thanks!

PR #158, opened 2016-10-18
Fixes #141: Make a more detailed environment list.

### LINKED ISSUE #141: Make stack lister more useful.
Things stack lister should do:
- Actually care about tagged values, not just name matching (see #99, should use this tag)
- Show creation date/time, whatever other interesting values are available in a nicely drawn table (like status, and other commands)
- Sort / filter values?

askreet:
This is a first pass here, we could improve the view more, filter by status, colorize some values, etc.


askreet:
Looks like this:

PR #161, opened 2016-10-22
Add interactive prompting for missing parameters on create/update.

askreet:
Some of the improvements in this patch have lead me to a really simple implementation of a really complicated internal plugin we had that used KMS Keys to encrypt parameter values. I'll be putting up another patch to add this feature to core, as an example of how we can use the parameter collections and plugins as a sort of middleware model.

### DIFF
--- a/lib/moonshot.rb
+++ b/lib/moonshot.rb
@@ -56,6 +56,8 @@ end
   'stack_events_poller',
   'merge_strategy',
   'default_strategy',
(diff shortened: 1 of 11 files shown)


askreet:
Ping @pdrakeweb @glennpratt @borsothy, would love to get some eyes on this patch if anyone has a few spare moments. :)

glennpratt:
:+1:

PR #167, opened 2016-11-14
Fix the usege of MOONSHOT_SSH_USER and MOONSHOT_SSH_KEY_FILE environment variables

ProZsolt:
These variables not working after 0.7.7


askreet:
Thanks, @ProZsolt -- my bad!

Issue #169, opened 2016-11-15
Moonshot 1.0 tag changes cannot work with some resource types.


askreet:
In particular, DBCluster tags can never be updated. We need to implement a config option for a tagging strategy so that stacks with these resources already launched in production can continue to use the same tagging strategy.

I would imagine it looks something like two classes called `LegacyTaggingStrategy` and `TaggingStrategy`. Each would be set in the `ControllerConfig`, defaulting to `TaggingStrategy`. The `create`, `update` and `list` actions would use the strategy to understand how to create, update and find resources, respectively.

```
# Example Moonfile.rb file using the legacy strategy
Moonshot.config do |c|
  c.tagging_strategy = Moonshot::LegacyTaggingStrategy.new
end
```

askreet:
We should probably include a note in the 1.x cutover guide as well about why you might choose to use the legacy strategy.

PR #170, opened 2016-11-16
Tag Update fixed and made list aware of old stages. #169

Skarlso:


askreet:
Tested this locally -- thanks for the patch!

Skarlso:
Huzzah!

PR #172, opened 2016-11-21
Removes light dark text coloring that is invisible in terminals configured with solarized dark

nickjanus:
This caused me some consternation when I saw:
```
 ❮ bundle exec moonshot update --no-interactive-logger --show-all-events -n njanus
                 :
```
Since this stack parameter had no description, I had no idea what parameter moonshot was asking me to input. Then I highlighted the prompt and discovered that moonshot printed the parameter name.

### DIFF
--- a/lib/moonshot/ask_user_source.rb
+++ b/lib/moonshot/ask_user_source.rb
@@ -29,9 +29,9 @@ module Moonshot
     private
 
     def prompt
-      print "(#{@sp.name})".light_black
+      print "(#{@sp.name})"
       print " #{@sp.description}" unless @sp.description.empty?
-      print " [#{@sp.default}]".light_black if @sp.default?
+      print " [#{@sp.default}]" if @sp.default?
       print ': '
     end
   end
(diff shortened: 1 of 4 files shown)


askreet:
I'm having a hard time reproducing this, any chance your terminal isn't configured properly?

I'm using iTerm2 on OSX with Solarized Dark pre-defined color set to test, and I wrote this Ruby program:

```ruby
#!/bin/env ruby
require 'colorize'

colors = %i(black light_black white light_white)
colors.each do |c|
  puts "I am #{c} text.".send(c)
end
```

When I run it, I see:

nickjanus:
This could very well be a config issue on my end.  After reloading the solarized defaults:
image
I'm using zsh, but switching to bash doesn't seem to help.  This could very well be a config issue, so I'm not concerned about this getting merged.

nickjanus:
Oh hai, iterm2 settings.  It looks like turning up minimum contrast works around the coloring issue so... meh to this patch?

image

askreet:
@nickjanus Hmm, any chance you turned that down? I don't think I've ever touched it and mine is at ~40%.

PR #173, opened 2016-11-22
Allow optional arguments for commands

Yitsushi:
This patch allows us to use optional arguments on `execute` like this one:

```
moonshot reset [version] [options]

moonshot reset                   # for development version
moonshot reset my_custom_version # for custom version
```

### DIFF
--- a/lib/moonshot/command_line_dispatcher.rb
+++ b/lib/moonshot/command_line_dispatcher.rb
@@ -16,7 +16,8 @@ module Moonshot
       parser = build_parser(handler)
       parser.parse!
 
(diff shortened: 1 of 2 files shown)


askreet:
This seems like a good change to me. Could you add a couple of RSpec tests that assert this works against a mock subclass of Moonshot::Command?

Yitsushi:
@askreet : A simple spec added. I don't know yet if I need more or not. Let me know if you think I have to extend this or other specs also.

askreet:
@Yitsushi LGTM.

PR #174, opened 2016-11-22
Allow custom configuration through Moonfile.rb with config.custom

Yitsushi:
This patch allows us to add custom configuration entries if custom commands wants to use them in general like "Is this a production stack?"

```ruby
## Moonfile.rb

production_accounts = ['dev-aws-account', 'dev2-aws-account']

Moonshot.config do |c|
  current_aws_account = Aws::IAM::Client.new.list_account_aliases.account_aliases.first
  c.custom['is_production'] = production_accounts.include?(current_aws_account)
end

## my_command.rb

class MyCommand << Moonshot::Command
  self.description = 'Do something dangerous.'
  def execute
    if controller.config.custom['is_production']
      raise "This command is not allowed in production environment" 
    end
    do_something
  end
end
```

### DIFF
--- a/lib/moonshot/controller_config.rb
+++ b/lib/moonshot/controller_config.rb
@@ -28,6 +28,8 @@ module Moonshot
     attr_accessor :ssh_command
     attr_accessor :ssh_config
     attr_accessor :ssh_instance
+    # Custom configuration for custom commands
+    attr_accessor :custom
 
     def initialize
       @default_parameter_source = AskUserSource.new
@@ -41,6 +43,7 @@ module Moonshot
       @project_root             = Dir.pwd
       @show_all_stack_events    = false
       @ssh_config               = SSHConfig.new
+      @custom                   = {}
 
       @dev_build_name_proc = lambda do |c|
         ['dev', c.app_name, c.environment_name, Time.now.to_i].join('/')


askreet:
I wonder if we should start treating the account alias you're running the command in as a first-class configuration component. I know we do some conditional logic on a few projects internally that could be replaced with something like:

```ruby
# Moonfile.rb
Moonshot.config do |c|
  c.artifact_repository = S3Bucket.new('my-project-builds-dev')

  c.in_account 'production-services' do |c|
    c.artifact_repository.bucket_name = 'my-project-builds-promoted'
  end
end

# moonshot/cli_extensions/dangerous_command.rb
class DangerousCommand < Moonshot::Command
  self.only_in_accounts = [ 'development-services' ]
  self.description = 'Load in some fake data!'

  # Please don't ever run this in production...
  def execute
    # ...
  end
end

Yitsushi:
`in_account` logic is ok.

Custom configuration maybe can helpful outside of the account topic. If you want to manage some extra value for multiple command, like restrict or extend the command functionality depends on the environment. For example `bucket-prefix` for S3 custom commands.

Of course we can handle it with `include` but if it's only a single value I think a while module/class would be a bit over-engineering. Now configuration does not allow to make custom properties. If a company wants to use a few configuration options that does not exists in moonshot `ControllerConfig` as `attr_accessor`, than they have to modify moonshot or find a hacky way that maybe painful after a few commit/refactor.

Of course you can always find a workaround for every single problem because there are more solution for one problem.

`in_account` is a solution for the problem with separated logic between AWS accounts, but (and I can't list other use-cases) there could be more than just account specific logic.

askreet:
I get what you're saying, but we have one use case right now. I worry about implementing a generic bucket for custom configuration, as it might not force the conversation around missing core functionality. Imagine we have 3 or 4 plugins all using this custom hash in different ways, with no safety or consistency provided by Moonshot in a way that might be generically useful to lots of projects.

I'd rather address the issue of account-specific controller configuration and then see if a use case arises for custom configurations outside of that.

I also think we could do something where plugins can register configuration options, so that you get the guardrails against setting a configuration option you expect to work, but does not if you e.g. have not included the right plugin source.

Yitsushi:
Agree.

PR #192, opened 2016-12-31
Set sample to use current moonshot version.

tomelliot16:
Seems the version should be bumped


tomelliot16:
@askreet Has a Moonfile but no Moonfile.rb. It should probably be added with this patch. It's in the same format right or did it change?

tomelliot16:
@askreet lol travis is made at your naming convention 😝 Makefile.rb is not in snake case. Should we add it to the ignore list?

askreet:
@tomelliot16 Yeah, I think so.

Issue #193, opened 2017-01-11
Rollout Launch Configuration should consider ASG:Missing/ELB:Missing status


Skarlso:
Currently, rollout launch configuration does not consider ASG:Missing/ELB:Missing status, thus hangs forever.

Skarlso:
Done.

PR #194, opened 2017-01-11
Issue #193 | Implementing abort on instance gone missing.

Skarlso:


lib/moonshot/tools/asg_rollout.rb
```diff
@@ -65,6 +65,10 @@ def wait_for_in_service(new_instance)
 
           loop do
             instance_health = asg.instance_health(new_instance)
+            if instance_health.out_of_service?
```
askreet: The instance will always start in OutOfService, before it (hopefully) moves into InService. What we might want to do here is implement a timeout, and if we've passed that timeout and we're in either OutOfService state *or* missing (which `out_of_service?` may cover?), then abort.

Skarlso: No it doesn't. It starts in Pending. So this is correct. :)

askreet: Oh, I misunderstood this as being ELB health. Maybe this is OK -- let me do some local testing when I have a chance.

askreet: Tested this locally, does what it says on the tin:
```
[ \ ] [ 0m 40s ] Waiting for a new instance to join Auto Scaling Group...
[ ✗ ] [ 0m 37s ] Error while performing step: Waiting for i-03f606ef9face4526 to be InService...
  RuntimeError: Instance i-03f606ef9face4526 went OutOfService while waiting to join...
[ ✓ ] [ 0m 0s ] Restored MaxSize/DesiredCapacity values to normal!
```

However, I think we should just try again with the next instance that joins. If the user sees this happen 2-3 times they might give up and press C-c, which we'll handle. Thoughts?

Skarlso: I'm not entirely against that. I just fear if people trigger it, and than leave it there and wait for it to finish. And then forget about it, come back to it to see it still running after 30, or so minutes. 

But thinking about this as I type, it's basically the same if they leave it now, it would just quit. So.. Yeah. We can do that. But is there merit in it? Like.. do you think if the first one wasn't able to attach the next one would because the problem the first one saw would go away? 

I mean, when I was testing this, it couldn't create a new instance because we exceeded volumes. So any next one which would have tried to join, wouldn't have succeeded either. 

I'm just brainstorming here. :)

Skarlso:
@askreet Update the code. With this behavior, if the new instance dies, it tries again. Indefinitely for now. 

```
[ ✓ ] [ 0m 3s ] Increased MaxSize/DesiredCapacity by 1.
[ ✓ ] [ 0m 32s ] A wild i-0f49ce35a3063f962 appears!
[ ✗ ] [ 0m 30s ] Error while performing step: Waiting for i-0f49ce35a3063f962 to be InService...
  RuntimeError: Instance i-0f49ce35a3063f962 went OutOfService while waiting to join...
Instance i-0f49ce35a3063f962 went OutOfService while waiting to join...
[ ✓ ] [ 1m 28s ] A wild i-01c798a5fc69e8340 appears!
[ ✓ ] [ 3m 7s ] Instance i-01c798a5fc69e8340 is ASG:InService/ELB:InService!
[ ✓ ] [ 0m 1s ] Restored MaxSize/DesiredCapacity values to normal!
```

PR #195, opened 2017-01-18
Updated Backup plugin to work with Moonshot 1.x.

borsothy:


borsothy:
@askreet if this could make it into 1.1.0, that would be great.

borsothy:
@askreet just one more question: how come [this](https://github.com/acquia/moonshot/commit/ed3e624d84c22ca69ceb3a3b7630771e4209dbf5#diff-28e0aed7c6ca653f90572434ad48f64eR74) is not a major update? The controller is injected into the plugins methods, therefore some might break with this modification. Is it possible to add default values for those params?

askreet:
@borsothy I think a major version update is probably right. I'll call the next beta 2.0.0.beta1.

borsothy:
@askreet @jamesiarmes comment addressed.

PR #196, opened 2017-01-27
EncryptedParameters plugin should be in the Moonshot::Plugins module

janost:
Refactoring the `EncryptedParameters` plugin to be in the `Moonshot::Plugins` module instead of `Moonshot` to conform with its own documentation:
https://github.com/acquia/moonshot/blob/master/lib/plugins/encrypted_parameters.rb#L14


askreet:
Yup, good catch. Fortunately we're still in beta releases, so I think breaking this is OK.

PR #197, opened 2017-01-27
Add ruby 2.3 to .travis.yml

janost:
We are using `moonshot` with ruby 2.3 and I suspect a lot of people are running it on 2.3, since Ubuntu 16.04 LTS ships with it by default.
It seems to work fine and tests are running fine, so it might be good to have travis run tests with this version.

Fixed error on travis: `can't modify frozen String`:
https://github.com/bundler/bundler/issues/5357


askreet:
LGTM, thanks!

PR #205, opened 2017-02-20
Fix status command if there's a standby instance in the ASG

janost:
This fixes an issue when there's a standby instance in one of the ASGs.

```
$ moonshot status
┌─ CodeDeploy Application: app-dev-user
│
│ Application and Deployment Group are configured correctly.
│
└──
CloudFormation Stack app-dev-user exists.
undefined method `public_ip_address' for nil:NilClass
Did you mean?  public_send (at /home/user/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/bundler/gems/moonshot-494af420cef0/lib/moonshot/stack_asg_printer.rb:117:in `instance_row')
```

### DIFF
--- a/lib/moonshot/stack_asg_printer.rb
+++ b/lib/moonshot/stack_asg_printer.rb
@@ -111,19 +111,18 @@ module Moonshot
     end
 
     def instance_row(asg_instance, ec2_instance)
-      public_ip = if ec2_instance
-                    if ec2_instance.public_ip_address
-                      ec2_instance.public_ip_address
-                    else
-                      "#{ec2_instance.private_ip_address} (PRV)"
-                    end
-                  else
-                    'offline'
-                  end
-      uptime = ec2_instance.nil? ? 'offline' : uptime_format(ec2_instance.launch_time)
+      if ec2_instance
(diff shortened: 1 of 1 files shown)


lib/moonshot/stack_asg_printer.rb
```diff
@@ -111,13 +111,22 @@ def create_instance_table(asg_info)
     end
 
     def instance_row(asg_instance, ec2_instance)
+      public_ip = if ec2_instance
```
askreet: Some thoughts:
- Rename 'public_ip', as it's not always public now.
- Leave a comment about why we think an instance might not be responding in EC2 (others have seen this too).
- I think 'offline' sounds too much like an understood condition, when really this appears during weird "eventually consistent" states at AWS. I'm thinking 'unknown' instead?

Maybe something like:

```ruby
if ec2_instance
  ip_address = ec2_instance.public_ip_address || "#{ec2_instance.private_ip_address} (PRV)"
  update = uptime_format(ec2_instance)
else
  # We've seen race conditions where ASG tells us about instances that EC2 is no longer
  # aware of.
  ip_address = 'unknown'
  uptime = 'unknown'
end

janost:
Thanks @askreet, I've implemented some changes based on your feedback.

PR #206, opened 2017-02-20
Fix ssh command when using multiple ASGs

janost:
`moonshot ssh` wasn't working properly in case of multiple ASGs.


askreet:
Nice catch, thanks!

PR #207, opened 2017-02-20
Added support for optional ASGs in CodeDeploy

janost:
This feature allows the user to specify an array of optional ASGs for `Moonshot::DeploymentMechanism::CodeDeploy`.
The application will be deployed to the ASGs listed here the same way as the `asg` parameter, but the deployment won't fail if the optional ASGs don't exist, `moonshot` will simply skip them.

We use this feature internally in our app, but it could be useful for someone else.

### DIFF
--- a/docs/mechanisms/deployment.md
+++ b/docs/mechanisms/deployment.md
@@ -35,9 +35,9 @@ manage a Deployment Group for in CodeDeploy.
 
 ### optional_asg | string,array
 
-The logical name of one or more Auto Scaling Groups to create and
-manage a Deployment Group for in CodeDeploy. This ASG doesn't have
-to exist. If it does, it will be added to the Deployment Group.
+The logical name of one or more Auto Scaling Groups to add to the
+Deployment Group in CodeDeploy. These ASGs don't have to exist.
+If they do, they will be added to the Deployment Group.
 
 ### role | string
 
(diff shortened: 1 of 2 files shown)


lib/moonshot/deployment_mechanism/code_deploy.rb
```diff
@@ -29,13 +33,16 @@ class Moonshot::DeploymentMechanism::CodeDeploy # rubocop:disable ClassLength
...
+  # rubocop:disable Metrics/ParameterLists
   def initialize(
       asg: [],
+      optional_asg: [],
       role: 'CodeDeployRole',
       app_name: nil,
       group_name: nil,
       config_name: 'CodeDeployDefault.OneAtATime')
-    @asg_logical_ids = asg.is_a?(Array) ? asg : [asg]
+    @asg_logical_ids = [*asg]
```
askreet: The idiomatic Ruby syntax for this is:

```ruby
@asg_logical_ids = Array(asg)
```

docs/mechanisms/deployment.md
```diff
@@ -33,6 +33,12 @@ Parameters
 The logical name of one or more Auto Scaling Groups to create and
 manage a Deployment Group for in CodeDeploy.
 
+### optional_asg | string,array
+
+The logical name of one or more Auto Scaling Groups to create and
```
askreet: This first sentence sounds a bit like it creates a new DG for the optional ASGs, but in fact (if I understand the code correctly) all ASGs are added to the same DG.

PR #208, opened 2017-02-22
Parent stacks command only allowing one parent

jfarrell:
Bug in the option parser where it was only pushing one
parent stack from the command line even through the config
loader would later loop over all parents provided.

### DIFF
--- a/lib/moonshot/commands/parent_stack_option.rb
+++ b/lib/moonshot/commands/parent_stack_option.rb
@@ -8,6 +8,11 @@ module Moonshot
                   'Parent stack to import parameters from') do |v|
           Moonshot.config.parent_stacks = [v]
         end
+
+        parser.on('--parents a,b,c', Array,
+                  'List of parent stacks to import parameters from') do |v|
+          Moonshot.config.parent_stacks = v
+        end
       end
     end
   end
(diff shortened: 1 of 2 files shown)


lib/moonshot/commands/parent_stack_option.rb
```diff
@@ -6,7 +6,7 @@ def parser
 
         parser.on('-pPARENT_STACK', '--parent=PARENT_STACK',
                   'Parent stack to import parameters from') do |v|
-          Moonshot.config.parent_stacks = [v]
+          (Moonshot.config.parent_stacks ||= []).push(v)
```
askreet: So this is a subtle change in behavior that might impact users, actually. Imagine you have a default value set in the Moonfile.rb of `['default-parent-stack-name']`. With this change, adding `-p stack-name` would set it to `['default-parent-stack-name', 'stack-name']`.

I'm not sure which is "right". I tend to lean towards wanting the presence of command-line options to replace whatever is already set.

One hacky way I could imagine doing this would be with an instance variable, though I hate it already. Something like:

```ruby
parser.on('-pPARENT_STACK', '--parent=PARENT_STACK', '...') do |v|
  Moonshot.config.parent_stacks = [] unless @_reset_cli_parent_stacks
  @_reset_cli_parent_stacks = true
  Moonshot.config.parent_stacks << v
end
```

Maybe a less hacky option would be storing the values passed here into an instance variable and then having a post-parse callback to replace the current config option with the value if it's non-empty.

I think we can write tests asserting the current behavior and the intent of this change by setting up `Moonshot.config.parent_stacks` before calling the parser both with and without a default value.  IIRC the `Moonshot.config` is reset before each example.

jfarrell:
@askreet PR updated to use new --parents flag to override and not impact existing functionality of --parent flag

askreet:
Looks good, thanks!

PR #212, opened 2017-03-16
new --refresh-parameters flag for update

jfarrell:
Adds a new --refresh-parameters flag for update calls allowing stacks
to refresh parent stack parameters if necessary


askreet:
Tested this locally, nice work. We'll have to do another major bump at next cut (since this adds a parameter to `#update`.)

PR #213, opened 2017-03-20
Added support for changing changeset wait time

janost:
AWS sometimes fails to create a changeset in 30 seconds when we're running a `moonshot update`, causing our release jobs to fail.

The following has been implemented in this PR:
- Changeset wait time is now configurable, defaults to 90 seconds.
```moonfile.rb
Moonshot.config do |c|
  ...
  c.changeset_wait_time = 120
end
```
- Slowed down science a bit since [other waiters seem to wait at least 5 seconds](https://github.com/aws/aws-sdk-ruby/blob/master/aws-sdk-core/apis/cloudformation/2010-05-15/waiters-2.json#L5) before repeating requests.

### DIFF
--- a/lib/moonshot/change_set.rb
+++ b/lib/moonshot/change_set.rb
@@ -76,7 +76,10 @@ module Moonshot
     # NOTE: At the time of this patch, AWS-SDK native Waiters do not
     # have support for ChangeSets. Once they add this, we can make
     # this code much better.
+    # Still no support for this waiter, but it's planned.
+    # https://github.com/aws/aws-sdk-ruby/issues/1388
     def wait_for_change_set
+      wait_seconds = Moonshot.config.changeset_wait_time || 30
       start = Time.now.to_i
 
       loop do
@@ -89,11 +92,11 @@ module Moonshot
           return
         end
 
-        if Time.now.to_i > start + 30
-          raise 'ChangeSet did not complete creation within 30 seconds!'
+        if Time.now.to_i > start + wait_seconds
+          raise "ChangeSet did not complete creation within #{wait_seconds} seconds!"
         end
 
-        sleep 0.25 # http://bit.ly/1qY1ZXJ
+        sleep 5 # http://bit.ly/1qY1ZXJ
       end
     end
   end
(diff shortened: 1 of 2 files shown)


smatyas:
This makes our release process fail a lot of times. Please review and merge it ASAP. Thanks!

ghaabor:
+1 on this

lib/moonshot/change_set.rb
```diff
@@ -89,11 +92,11 @@ def wait_for_change_set
...
         end
 
-        if Time.now.to_i > start + 30
-          raise 'ChangeSet did not complete creation within 30 seconds!'
+        if Time.now.to_i > start + wait_seconds
+          raise "ChangeSet did not complete creation within #{wait_seconds} seconds!"
         end
 
-        sleep 0.25 # http://bit.ly/1qY1ZXJ
+        sleep 5 # http://bit.ly/1qY1ZXJ
```
glennpratt: This seems a bit much to apply across the board, but I'm not familiar with the issue

glennpratt: Maybe just put the comment from the PR description here.

janost: I explained the reason for this in the PR. It seems other waiters provided by aws-sdk has at least a 5 seconds delay to avoid spamming the AWS API. We can revert this change if this delay is not acceptable.

janost: I've added a comment about the 5 seconds delay.

janost:
This still often causes headaches for us. Is it possible to discuss this further and maybe merge it?

lib/moonshot/change_set.rb
```diff
@@ -76,7 +76,10 @@ def delete
     # NOTE: At the time of this patch, AWS-SDK native Waiters do not
     # have support for ChangeSets. Once they add this, we can make
     # this code much better.
+    # Still no support for this waiter, but it's planned.
+    # https://github.com/aws/aws-sdk-ruby/issues/1388
     def wait_for_change_set
+      wait_seconds = Moonshot.config.changeset_wait_time || 30
```
pdrakeweb: Any reason not to increase the default here?

janost: The reason I decided to keep 30 seconds as the default is because I suspect the changeset creation time depends on the size of the template and I'm not sure how many users are affected by this issue.

pdrakeweb: Makes sense.  I see it in the 20+ second range from time to time.  I feel like we should cover at least the 95% case here so 60 or 90 seconds seems reasonable.

janost: Alright, I'll increase it to 90 seconds.

PR #214, opened 2017-03-23
CodeDeploySetup plugin

jfarrell:
New plugin to auto-create any S3 Buckets to be used by CodeDeploy. Also
updates CodeDeploy deploy mechanism to create the IAM Role necessary to
successfully deploy code.

### DIFF
--- a/docs/plugins/code_deploy_setup.md
+++ b/docs/plugins/code_deploy_setup.md
@@ -1,0 +1,60 @@
+# Moonshot CodeDeploy Setup plugin
+Moonshot plugin to ensure CodeDeploy has all the necessary S3 buckets created
+ to support deploying to multiple configured regions.
+
+## Moonshot configuration
+Update your Moonfile.rb with the following configuration
+
+```ruby
+Moonshot.config do |c|
+  # ...
+  c.app_name = 'example-project'
+
+  bucket_name = "#{c.app_name}-builds"
+
+  # bucket_name: Base name of the S3 bucket to create, will have region appended to it.
+  # prefix: Optional. If using the same bucket for multiple apps then recommend adding a prefix.
+  # regions: list of all supported regions you would like to use. Default: ENV['AWS_REGION'].
+  code_deploy_setup_plugin = Moonshot::Plugins::CodeDeploySetup.new(
+    bucket_name,
+    prefix: c.app_name,
+    regions: [ 'us-east-1', 'us-west-2' ]
+  c.plugins << code_deploy_setup_plugin
+
+  c.deployment_mechanism = CodeDeploy.new
+  c.artifact_repository = S3BucketViaGithubReleases.new(
+    code_deploy_setup_plugin.bucket_name,
+    prefix: code_deploy_setup_plugin.bucket_prefix
+  )
+```
+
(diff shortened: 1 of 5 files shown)


jfarrell:
@glennpratt if you have time can I please get a review/merge on this PR

lib/plugins/code_deploy_setup.rb
```diff
@@ -0,0 +1,57 @@
...
+
+      # Hook entry points to ensure S3 Buckets are available for CodeDeploy
+      def run_hook(_resources)
+        setup_code_deploy_s3_buckets
+      end
+
+      # Moonshot hooks to trigger this plugin
+      alias pre_create run_hook
+      alias pre_deploy run_hook
+      alias pre_doctor run_hook
```
glennpratt: I'm not sure how I feel about this one.

Does doctor have other places where it actually does work?

I'd probably drop this and just make sure the doctor output tells you that this will likely resolve itself.

docs/plugins/code_deploy_setup.md
```diff
@@ -0,0 +1,60 @@
...
+
+  bucket_name = "#{c.app_name}-builds"
+
+  # bucket_name: Base name of the S3 bucket to create, will have region appended to it
+  # prefix: Optional. If using the same bucket for multiple apps then recommend adding a prefix
+  # regions: list of all supported regions you would like to use. Default: ENV['AWS_REGION']
+  code_deploy_setup_plugin = Moonshot::Plugins::CodeDeploySetup.new(
+    bucket_name,
+    prefix: c.app_name,
+    regions: [ 'us-east-1', 'us-west-2' ]
```
jfarrell: Thought was that the given app would define the specific regions it ran in, CDB has an AWS helper that has the list of their specific supported regions that get passed into this plugin

lib/plugins/code_deploy_setup.rb
jfarrell: doctor would call the role check before and then fail if the role was missing, with this change it now will create the role and allow doctor to pass if it is successfully created

docs/plugins/code_deploy_setup.md
glennpratt: I see it has a constant, I guess that could change to something that could look up the current, global AWS account and map regions as needed.

docs/plugins/code_deploy_setup.md
```diff
@@ -0,0 +1,60 @@
...
+```ruby
+Moonshot.config do |c|
+  # ...
+  c.app_name = 'example-project'
+
+  bucket_name = "#{c.app_name}-builds"
+
+  # bucket_name: Base name of the S3 bucket to create, will have region appended to it
+  # prefix: Optional. If using the same bucket for multiple apps then recommend adding a prefix
+  # regions: list of all supported regions you would like to use. Default: ENV['AWS_REGION']
```
jfarrell: will add

lib/plugins/code_deploy_setup.rb
jfarrell: this plugin probably does not need a pre_doctor since the deployment_mechanism will cover that check, will remove

PR #215, opened 2017-04-04
Wait for tag to exist on github in github release build mechanism.

kaushik:


lib/moonshot/build_mechanism/github_release.rb
```diff
@@ -140,6 +143,20 @@ def hub_release_exists(semver)
       false
     end
 
+    def hub_find_remote_tag(tag_name)
+      attempts = 0
+      # Attempt to find the build. Rescue and re-attempt if the build can not
+      # be found on github yet.
+      begin
+        sleep 1 # Give some time for tag to showup.
```
pdrakeweb: Should we move this into the rescue so that we only wait for the tag if it isn't there on the first attempt?

lib/moonshot/build_mechanism/github_release.rb
```diff
@@ -140,6 +143,20 @@ def hub_release_exists(semver)
...
+        sh_out("hub ls-remote --exit-code --tags upstream #{tag_name}")
+      rescue RuntimeError => e
+        attempts += 1
+        retry unless attempts >= MAX_GITHUB_TAG_FIND_ATTEMPTS
```
jamesiarmes: This should probably be configurable and we may want to sleep for longer than a second. On multiple occasions I've seen GitHub take more than a minute to recognize something that's been pushed to it.

lib/moonshot/build_mechanism/github_release.rb
kaushik: Yeah, we can do that. I put it here because the first attempt is definitely bound to fail. We can accept that cos it will retry or not remove sleep from line 116

pdrakeweb: So, we know that github fails for N seconds all the time?  If so, this sleep should be at least N (above, this sleep was 2).  I'm fine with this being here but there should probably be a comment that we sleep first because we know that this always fails for the first N seconds.

lib/moonshot/build_mechanism/github_release.rb
```diff
@@ -140,6 +141,22 @@ def hub_release_exists(semver)
...
     end
 
+    def hub_find_remote_tag(tag_name)
+      tries = 1
+      # Attempt to find the build. Rescue and re-attempt if the build can not
+      # be found on github yet.
+      begin
+        # Start with a sleep because the tag generally takes about a second
+        # to show up on github.
+        sleep tries * tries
```
borsothy: I agree with the exponential retry, but with the default retry value this can take up to 20 mins. Is that okay? I doubt it will take that long for github to realise there was a tag created.

borsothy:
The commonmarker update is causing problems on Travis too.

lib/moonshot/build_mechanism/github_release.rb
kaushik: 225/60 ~ 4 mins. Did I do my math wrong?

kaushik:
Travis is failing. I ran rake locally and here is the report,
```
 ± |CPD-1574 ✓| → be rake
/Users/kaushik.sirineni/.rvm/rubies/ruby-2.2.3/bin/ruby -I/Users/kaushik.sirineni/.rvm/gems/ruby-2.2.3/gems/rspec-core-3.4.4/lib:/Users/kaushik.sirineni/.rvm/gems/ruby-2.2.3/gems/rspec-support-3.4.1/lib /Users/kaushik.sirineni/.rvm/gems/ruby-2.2.3/gems/rspec-core-3.4.4/exe/rspec --pattern spec/\*\*\{,/\*/\*\*\}/\*_spec.rb
rspec pid: 86840

Moonshot::Tools::ASGRollout
  for a basic configuration
    should perform the normal procedure
  when a TerminateWhen hook does not complete in the timeout
    should fail
  when the pre-detach hook returns false
    should abort with a failure before detaching the instance

Moonshot::BuildMechanism::GithubRelease
  #doctor_hook
    should call our hooks
    #doctor_check_upstream
      should fail without upstream.
      should succeed with upstream remote.
    #doctor_check_hub_auth
      should succeed with 0 exit status.
      should critical with non-zero exit status.
  #git_tag_exists
I, [2017-04-05T12:27:56.333160 #86840]  INFO -- : tag 0.0.0-rspec already exists
    passes if the tag exists at the same sha
    fails if the tag does not exist
    errors if the tag exists at a different sha
  #hub_release_exists
I, [2017-04-05T12:27:56.335376 #86840]  INFO -- : release 0.0.0-rspec already exists
    passes if the github release exists
    fails if the github release does not exist

Moonshot::BuildMechanism::TravisDeploy
  #doctor_hook
    should call our hooks
    #doctor_check_travis_auth
      should pass if travis exits 0
      should pass fail travis exits 1
  #wait_for_build
    should only make one attempt if the job is found
    should only make the max number of attempts before failing
    should make attempts until the build is found
  #wait_for_job
    should continue until the job is complete
    should fail if the job does not complete within the time limit

optional argument for command
  should let us run OptionalArgumentCommand without extra arguments
  should let us run OptionalArgumentCommand with extra argument

Moonshot::Command
  #parser
    should parse the verbose option

Moonshot::Commands::Console
  #execute
    should start a pry session

Moonshot::Commands::Create
  Should process ["-P", "Key=Value", "-P", "OtherKey=OtherValue"] correctly
  Should process ["-PKey=ValueWith=Equals"] correctly
  should handle version and deploy correctly
  should process multiple parent stacks

Moonshot::Commands::Update
  Should process ["-P", "Key=Value", "-P", "OtherKey=OtherValue"] correctly
  Should process ["-PKey=ValueWith=Equals"] correctly

create
  when parent, answer and overrides are all provided
Imported 2 parameters from parent stack parent-stack-1!
    use parameter precedent correctly.
  when some parameters are missing, but have no defaults
Imported 2 parameters from parent stack parent-stack-1!
    should raise an error

update
  when a new parameter is added to the stack with a template
    use parameter precedent correctly.
  when refresh-parameters option is provided
    should preserve existing existing parent parameters
    should refresh all parent stack parameters

Moonshot::InteractiveLoggerProxy
  #debug?
    should be defined.
    should have method arguments compatible with InteractiveLogger#debug?
  #start
    should be defined.
    should have method arguments compatible with InteractiveLogger#start
  #start_threaded
    should be defined.
    should have method arguments compatible with InteractiveLogger#start_threaded
  #debug
    should be defined.
    should have method arguments compatible with InteractiveLogger#debug
  #info
    should be defined.
    should have method arguments compatible with InteractiveLogger#info
  #error
    should be defined.
    should have method arguments compatible with InteractiveLogger#error
  #msg
    should be defined.
    should have method arguments compatible with InteractiveLogger#msg

Moonshot::InteractiveLoggerProxy::Step
  #continue
    should be defined.
    should have method arguments compatible with InteractiveLogger::Step#continue
  #failure
    should be defined.
    should have method arguments compatible with InteractiveLogger::Step#failure
  #success
    should be defined.
    should have method arguments compatible with InteractiveLogger::Step#success
  #blank
    should be defined.
    should have method arguments compatible with InteractiveLogger::Step#blank
  #repaint
    should be defined.
    should have method arguments compatible with InteractiveLogger::Step#repaint

Moonshot::Plugins::Backup
  #new
    should yield self
    should raise ArgumentError if insufficient parameters are provided
    should raise ArgumentError if redundant parameters are provided
    should set a default value to target_name if not specified
  #to_backup
    should return a Backup object
    should raise ArgumentError when no bucket specified
    should set default config values
    should only respond to the default hooks
  #backup
    should raise ArgumentError if resources are not injected
    should return silent if account not found in buckets hash
    should upload

Moonshot::Plugins::CodeDeploySetup
  #new
    should raise ArgumentError if no region arg is set
    should use the ENV[AWS_REGION] if no region arg is set
    should use return the correct bucket name for the current active region
    should use return the correct bucket prefix
  #run_hook
    creates a bucket for all set regions

Plugins support
  calls defined methods on plugins in order, providing them with a Moonshot::Resources
  doesn't call an undefined method

Moonshot::Shell
  #shell
    should return a shell compatible with Thor::Shell::Basic.
  #sh_out
    should raise on non-zero exit.
    should not raise if fail is disabled.
  #sh_step
I, [2017-04-05T12:27:58.469986 #86840]  INFO -- : false
    should raise an error if the step fails.
I, [2017-04-05T12:27:58.480573 #86840]  INFO -- : echo yo
    should provide the step and sh output to a block.
I, [2017-04-05T12:27:58.496703 #86840]  INFO -- : echo 3otKXTS9JM2CbCW67u4miuH-GEBM9C2x0mGXolmw54uDwyPwiznLmwLv2yNyb4YF58xJV2IFXOwCpAsdjT4NOzv6g6VIyx1GPJieCL3bWoovqbdJb6TbRG3LVG8q...
    should truncate a long messages.

Moonshot SSH features
  Moonshot::Controller#ssh
    normally
      should execute an ssh command with proper parameters
    when an instance id is given
      should execute an ssh command with proper parameters

Moonshot::Stack
  #create
    when the stack creation takes too long
      should display a helpful error message and return false
    when the stack creation completes in the expected time frame
      should log the process and return true
    when the stack already exists
      should log a successful step and return true
    under normal circumstances
      should call CreateStack, then wait for completion
  #template_file
    should return the template file path
  #template
    should look for templates in the preferred order

Finished in 3.13 seconds (files took 3.32 seconds to load)
92 examples, 0 failures

Running RuboCop...
Inspecting 77 files
.............................................................................

77 files inspected, no offenses detected
```

PR #216, opened 2017-04-05
Wait for CI_status and release tag in github build mechanism

kaushik:


pdrakeweb:
We now have two separate retry implementations in these PRs.  I talked briefly with @askreet and he was amenable to introducing https://github.com/kamui/retriable.  We should refactor these two PRs to use the retriable gem.

kaushik:
Not just here, we have retry/wait in travis deploy as well, where retriable can be of help.

lib/moonshot/build_mechanism/github_release.rb
```diff
@@ -170,6 +169,27 @@ def fetch_changes(version)
...
 
+    # Checks for the commit's CI job status. If its not finished yet,
+    # wait till timeout.
+    #
+    # @param sha [String] Commit sha.
+    #
+    # @return [String] Status and links to the CI jobs
+    def check_ci_status(sha)
+      out = nil
+      ilog.start_threaded('Validate `CHANGELOG.md`.') do |step|
```
pdrakeweb: This doesn't appear to be validating changelog?

lib/moonshot/build_mechanism/github_release.rb
```diff
@@ -170,6 +169,27 @@ def fetch_changes(version)
...
+        Retriable.retriable on: RuntimeError,
```
pdrakeweb: Should we create a hash of default values and then merge in overrides for specific uses?

lib/moonshot/build_mechanism/github_release.rb
```diff
@@ -170,6 +169,27 @@ def fetch_changes(version)
...
+                            max_elapsed_timeout: @ci_status_timeout,
+                            multiplier: 0,
```
pdrakeweb: It seems like a multiplier of zero is going to make every interval zero (see https://github.com/kamui/retriable/blob/master/lib/retriable/exponential_backoff.rb#L19).

lib/moonshot/build_mechanism/github_release.rb
kaushik: Derp. Copied wrong text

lib/moonshot/build_mechanism/github_release.rb
kaushik: There are not many common values to share. Should find a nice location to put them in. I errd on defining them locally cos evry retry is kinda different in its own way.
But, I was thinking adding a retry option to [sh_out](https://github.com/kaushik/moonshot/blob/CPD-1834/lib/moonshot/shell.rb#L7) might be a good choice.
`
def sh_out(cmd, fail: true, stdin: '', retry: false, opts = {})
default_retry_optons.merge!(opts)
.
`

pdrakeweb: Retrying on RuntimeError instead of StandardError and having a reasonable multiplier seems like a good start.

askreet: Let's use a method to wrap these:

```ruby
def with_retries(&blk)
  default_options = {
    on: [ RuntimeError ],
    multiplier: 1.25
  }
  Retriable.retriable(default_options, &blk)
end
```

By the way, why are we retrying on `RuntimeError` instead of `StandardError`? Are we trying to allow a certain exception to pass through?

kaushik: Cos the sh_out raises RuntimeError on non zero exit. Anything else is not expected and we do not want to retry.

kaushik: btw are you guys talking about a default retry method to be used here or modifying sh_out in shell.rb to support retry as I suggested?

lib/moonshot/shell.rb
```diff
@@ -1,7 +1,17 @@
 require 'thor'
+require 'retriable'
 
 # Mixin providing the Thor::Shell methods and other shell execution helpers.
 module Moonshot::Shell
+  DEFAULT_RETRY_OPTIONS = {
+    on: RuntimeError,
+    tries: 50,
+    multiplier: 1,
+    base_interval: 1,
```
pdrakeweb: I feel like the default base interval should be a bit longer.  Retrying an execution every second for 5 minutes doesn't seem like the best default setting.

kaushik: How does 10 seconds sound?

lib/moonshot/shell.rb
```diff
@@ -1,7 +1,17 @@
 require 'thor'
+require 'retriable'
 
 # Mixin providing the Thor::Shell methods and other shell execution helpers.
 module Moonshot::Shell
+  DEFAULT_RETRY_OPTIONS = {
+    on: RuntimeError,
+    tries: 60,
+    multiplier: 1,
+    base_interval: 10,
```
kaushik: This is still 60 tries, should we further increase this or bring down the wait max period.

pdrakeweb: I would be tempted to say trying something every 10s for 2m is a good default.

kaushik: That seems reasonable to me. I will leave this comment as for sometime to see if other reviewers have an opinion on this.

lib/moonshot/shell.rb
```diff
@@ -51,4 +61,23 @@ def sh_step(cmd, args = {})
...
+  # @param opts [Hash] Options for retriable.
+  #
+  # @return [String] Stdout form the command.
+  def sh_retry(cmd, fail: true, stdin: '', opts: {})
+    Retriable.retriable(DEFAULT_RETRY_OPTIONS.merge(opts)) do
+      out = sh_out(cmd, stdin: stdin)
+      yield out if block_given?
+      out
+    end
+  rescue RuntimeError => e
```
borsothy: This is an interesting situation right here, cause even though `sh_out` can only raise `RuntimeError`, the function caller can raise _any_ exception from the block it has passed to the function.

PR #218, opened 2017-04-07
Improved error handling in 'download_from_github'

borsothy:


borsothy:
`file = Dir.glob("*#{version}*.tar.gz").fetch(0)` raises `IndexError` if the file was not found and therefore the download is not retried. Looking for feedback from @askreet @glennpratt @pdrakeweb @jamesiarmes

pdrakeweb:
This could probably stand to be done using Retriable as is being introduced by @kaushik to add some delay between retries.  Also, it seems worth changing that bit of code to not do a fetch on an empty array and instead begin the next Retriable iteration.

jamesiarmes:
Agree with @pdrakeweb. We should be checking to ensure the files were downloaded before trying to read them, rather than trying to catch an `IndexError` exception.

lib/moonshot/artifact_repository/s3_bucket_via_github_releases.rb
```diff
@@ -115,22 +116,22 @@ def upload_to_s3(file, key)
     # @raise [RuntimeError] If the file fails to download correctly after 3
     #                       attempts.
     def download_from_github(version)
+      filename = "*#{version}*.tar.gz"
       attempts = 0
-      begin
+
+      Retriable.retriable do
```
borsothy: Default configuration seemed reasonable to me.

lib/moonshot/artifact_repository/s3_bucket_via_github_releases.rb
```diff
@@ -115,22 +116,22 @@ def upload_to_s3(file, key)
...
         FileUtils.rm(Dir.glob('*'))
 
         # Download the release and find the actual build file.
         sh_out("hub release download #{version}")
-        file = Dir.glob("*#{version}*.tar.gz").fetch(0)
+        raise "File '#{filename}' not found." unless File.file?(filename)
 
+        file = Dir.glob(filename).fetch(0)
         unless (checksum = checksum_file(file)).nil?
           verify_download_checksum(file, checksum, attempt: attempts)
```
borsothy: I believe we also want to retry if the checksum verification has failed.

borsothy: I believe we also want to retry if the checksum verification has failed.

lib/moonshot/artifact_repository/s3_bucket_via_github_releases.rb
```diff
@@ -115,22 +116,22 @@ def upload_to_s3(file, key)
...
-      begin
+
+      Retriable.retriable on: RuntimeError do
         # Make sure the directory is empty before downloading the release.
         FileUtils.rm(Dir.glob('*'))
 
         # Download the release and find the actual build file.
         sh_out("hub release download #{version}")
-        file = Dir.glob("*#{version}*.tar.gz").fetch(0)
+        raise "File '#{filename}' not found." unless File.file?(filename)
```
pdrakeweb: file? is not documented as taking a pattern.  I presume you tested this and so it actually does?

lib/moonshot/artifact_repository/s3_bucket_via_github_releases.rb
```diff
@@ -115,22 +116,22 @@ def upload_to_s3(file, key)
     # @raise [RuntimeError] If the file fails to download correctly after 3
     #                       attempts.
     def download_from_github(version)
+      filename = "*#{version}*.tar.gz"
```
pdrakeweb: This should be named differently as this is not a filename, it is a pattern.

lib/moonshot/artifact_repository/s3_bucket_via_github_releases.rb
borsothy: No, you're right. Fixing

pdrakeweb: Given that this passed travis, then I think there are probably some missing unit tests as well.

borsothy:
@pdrakeweb I will add tests shortly.

borsothy:
I added some specs, they are passing and absolutely functional - however, I feel like the code is a bit ugly, I'd appreciate feedback on how to make this more clean.

lib/moonshot/artifact_repository/s3_bucket_via_github_releases.rb
```diff
@@ -115,22 +116,24 @@ def upload_to_s3(file, key)
     # @raise [RuntimeError] If the file fails to download correctly after 3
     #                       attempts.
     def download_from_github(version)
+      file = nil
```
jamesiarmes: I could be wrong, but I think `Retriable.retriable` will return the result of the block, so you could omit this line, and move the `file` at the end of this method to the last line of the block. I suggest giving this a try to verify.

borsothy: Nop, you're right. Addressing

PR #219, opened 2017-04-07
Stop printing error output when the release (expectedly) does not yet…

pdrakeweb:
… exist.


.travis.yml
```diff
@@ -7,6 +7,13 @@ cache: bundler
 # @see http://docs.travis-ci.com/user/workers/container-based-infrastructure/
 sudo: false
 
+addons:
```
pdrakeweb: This is a fix for the spec tests but the dependency was not introduced in this PR.

PR #220, opened 2017-04-18
Output the private IP regardless of an instance's public IP.

pingram3030:
For compliance, among many other, reasons it is best practice to use a jump host to access remote networks.

Currently the output of `moonshot status` only shows the external IP of an instance, when present, which is impossible to use for ssh, when on that private network. Outputting the private IP makes it easier for an admin to find and connect to a valid instance in an ASG.

Please let me know if you'd prefer a different format for the output :)


askreet:
Change looks good to me. At some point we will have to address how to format this output as it's quite long, but I don't think this change is the right time to start that discussion. Thanks!

PR #226, opened 2017-06-08
Make it possible to specify the backup bucket region.

pdrakeweb:
Buckets are specific to a region.  Currently, specifying a bucket in region A and then running moonshot in region B will fail to execute the backup.


pdrakeweb:
I have added a simple check that the configuration value is accepted and set, however, I did not add a test for its usage as it is used in a private method.

askreet:
👍  Thanks.

PR #231, opened 2017-08-15
Refactor paramater overriding. Set default parameter file

pingram3030:
So, I'm not 100% happy with this because it doesn't feel like the best writing of it, but it's been a while since I had to type out some ruby and some how I managed to fix a rubocop override! :D

We automatically load files from other folders inside of `moonshot/`, so why not also try and load a parameter file? If there is a better way of doing this I'd be welcoming of feedback, or if someone wants to Nike it that'd be fine also.

### DIFF
--- a/lib/moonshot/parameter_collection.rb
+++ b/lib/moonshot/parameter_collection.rb
@@ -17,6 +17,34 @@ module Moonshot
       obj
     end
 
+    def self.apply_overrides(config)
+      raise unless config.instance_of?(Moonshot::ControllerConfig)
+
+      default_answer_file = File.join(config.project_root,
+                                      'moonshot',
+                                      'params',
+                                      "#{config.environment_name}.yml")
+
+      answer_file = config.answer_file ? config.answer_file : default_answer_file
+
+      # The order we override parameters in is, in order of least to most important:
+      #   * project_root/moonshot/params/#{environment}.yml
+      #   * file defined by the user with -a/--answer-file
+      #   * individual parameters the user input with -P/--parameter
+      #
+      if File.readable?(answer_file)
+        YAML.load_file(answer_file).each do |key, value|
+          config.parameters[key] = value
+        end
+      end
+
+      config.parameter_overrides.each do |key, value|
+        config.parameters[key] = value
+      end
+
+      config
+    end
+
     def initialize
       @hash = {}
     end
--- a/lib/moonshot/controller.rb
+++ b/lib/moonshot/controller.rb
@@ -11,7 +11,7 @@ module Moonshot
       Moonshot::StackLister.new(@config.app_name).list
     end
 
-    def create # rubocop:disable AbcSize
+    def create
       # Scan the template for all required parameters and configure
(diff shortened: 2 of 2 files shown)


askreet:
I think the implementation looks good and I'm not against adding this feature since using it would be optional but I wonder what use cases you have for this. Prior to the big 2.x refactor we had environment configurations stored locally, and encouraged people to keep them committed to their source repo in cloud_formation/#{environment}.yml.

The goal of the 2.x refactor was to keep AWS as the "source of truth" when performing updates, so we merge the state of AWS into the UpdateStack call with any new changes (passed by -a/-P, etc.) -- why keep a local copy by the name of an environment?

I guess one use case I could see would be having a staging or QE environment that you routinely spin up/down and not having to pass `-a moonshot/staging.yml` for example.

jfarrell:
what about creating something like AlwaysUseDefaultSource, but for the answer_file. This way it would not break existing behavior, but for anyone wanting to use a default answer_file based on environment name they could set it in their Moonfile. We currently do something like this using a utility class

pingram3030:
How do you feel about me adding `param_dir` to `Moonfile.rb`, updating the method I made to suit and putting something in to `docs/user-guide/include_in_your_project.md`? I feel 'param_dir' is a bit more intuitive and means the user can put their per environment yaml where ever they please.

askreet:
@pingram3030 I actually prefer the by-convention approach of this patch to asking users to set it all up. Let's merge this patch in as-is, I just want to give it a manual test before I hit the green button.

If you want to add some docs showing how one might use this feature, that'd be great too.

lib/moonshot/parameter_collection.rb
```diff
@@ -17,6 +17,34 @@ def self.from_template(template)
...
+                                      'moonshot',
+                                      'params',
+                                      "#{config.environment_name}.yml")
+
+      answer_file = config.answer_file ? config.answer_file : default_answer_file
+
+      # The order we override parameters in is, in order of least to most important:
+      #   * project_root/moonshot/params/#{environment}.yml
+      #   * file defined by the user with -a/--answer-file
+      #   * individual parameters the user input with -P/--parameter
```
askreet: @pingram3030 Actually I was just thinking about this, does this mean that:

```
# 1. create stack with known param file containing 'FooParameter'
moonshot create -n staging

# 2. update FooParameter, -P takes precedence here
moonshot update -n staging -P FooParameter=newValue

# 3. update stack with new cloud formation template
moonshot update -n staging
```

I think that #3 would return the value of FooParameter to the value in the answer file the was changed by #2. If so, this breaks the pattern of allowing configuration change to live in AWS. We could change the algorithm to only use the by-name answer file for creation, and only use it for updates when a parameter is missing (such as a newly added parameter in an updated stack revision).

pingram3030:
I'll have it only check for the param file when creating. I'll definitely add to the doc for this, but I need time to do that; I've been hella starved of free time of late. Hopefully I can get back to this soon, it's something I'd like to have :)

PR #233, opened 2017-08-22
CPD-1298 | As a cloud data engineer, I would to able to skip the SHA status check when building a release.

hmagdy:
Testing Steps:
==

1. cd moonshot/lib/default
2. check --help without GithubRelease plugin

```
moonshot deploy --help 
Usage: moonshot deploy VERSION
    -v, --[no-]verbose               Show debug logging
    -n, --environment=NAME           Which environment to operate on.
        --[no-]interactive-logger    Enable or disable fancy logging
        --[no-]interactive           Use interactive prompts.
```
3. register github plugin by update Moonfile.rb with below content

```
Moonshot.config do |c|
  c.build_mechanism = GithubRelease.new('')
  c.deployment_mechanism = CodeDeploy.new(asg: 'AutoScalingGroup')
end
```
4- check --help again 
```
moonshot deploy --help 
Usage: moonshot deploy VERSION
    -v, --[no-]verbose               Show debug logging
    -n, --environment=NAME           Which environment to operate on.
        --[no-]interactive-logger    Enable or disable fancy logging
        --[no-]interactive           Use interactive prompts.
    -s, --skip-ci-status             Skips checks on CI jobs
        --ignore-app-stop-failures   Continue deployment on ApplicationStop failures
```

### DIFF
--- a/docs/mechanisms/build.md
+++ b/docs/mechanisms/build.md
@@ -33,6 +33,26 @@ Moonshot.config do |c|
 ...
 ```
 
+**skip_ci_status** is an optional flag. It would allow us to skip checks 
+on the commit's CI job status. Without this option, the GithubRelease mechanism will wait until the build is finished.
+
+Sample Usage
+
+```ruby
+Moonshot.config do |c|
+  wait_for_travis_mechanism = TravisDeploy.new("acquia/moonshot", true)
+  c.build_mechanism = GithubRelease.new(wait_for_travis_mechanism, skip_ci_status: true)
+...
+```
+Also a command-line option is available to override this value.
+
+	Usage: moonshot build VERSION
+	    -v, --[no-]verbose               Show debug logging
+	    -s, --skip-ci-status             Skip checks on CI jobs
+	    -n, --environment=NAME           Which environment to operate on.
+		--[no-]interactive-logger    Enable or disable fancy logging
+	    -F, --foo    
+
 ## TravisDeploy
 
 The Travis Build Mechanism waits for Travis-CI to finish building a
--- a/docs/user-guide/cli.md
+++ b/docs/user-guide/cli.md
@@ -16,6 +16,13 @@ Display debug logging. The types of things logged here are mostly
 useful only for core Moonshot developers, or if you are working on
 custom plugins.
 
+## `--skip-ci-status / -s`
+
+*Default: Disabled*
+
+It would allow us to skip checks on the commit's CI job status. 
+Without this option, the GithubRelease mechanism will wait until the build is finished.
+
 #### `--environment=NAME / -n NAME`
 
 *Default: dev-$USER, or as specified in the `environment_name`
--- a/lib/moonshot/build_mechanism/github_release.rb
+++ b/lib/moonshot/build_mechanism/github_release.rb
@@ -17,10 +17,20 @@ module Moonshot::BuildMechanism
     def_delegator :@build_mechanism, :output_file
 
     # @param build_mechanism Delegates building after GitHub release is created.
-    def initialize(build_mechanism, ci_status_timeout: 600, max_tag_find_timeout: 240)
+    def initialize(build_mechanism,
+                   ci_status_timeout: 600,
+                   max_tag_find_timeout: 240,
+                   skip_ci_status: false)
       @build_mechanism = build_mechanism
       @ci_status_timeout = ci_status_timeout
       @max_tag_find_timeout = max_tag_find_timeout
+      @skip_ci_status = skip_ci_status
+    end
+
+    def deploy_cli_hook(parser)
+      parser.on('-s', '--skip-ci-status', 'Skips checks on CI jobs', FalseClass) do |value|
+        @skip_ci_status = value
+      end
     end
 
     def doctor_hook
@@ -159,7 +169,7 @@ module Moonshot::BuildMechanism
       sh_step(cmd, msg: "Validate commit #{@sha}.") do |_, out|
         @commit_detail = out
       end
-      @ci_statuses = check_ci_status(@sha)
+      @ci_statuses = check_ci_status(@sha) if @skip_ci_status == false
     end
 
     def validate_changelog(version)
--- a/spec/moonshot/build_mechanism/github_release_spec.rb
+++ b/spec/moonshot/build_mechanism/github_release_spec.rb
@@ -189,5 +189,28 @@ module Moonshot # rubocop:disable ModuleLength
         expect { subject.send(:check_ci_status, sha) }.to raise_error(Shell::CommandError)
       end
     end
+    
+    describe '#validate_commit' do
+      let(:skip_ci) { false }
+      subject { 
+        s = described_class.new(build_mechanism, skip_ci_status: skip_ci);
+        s.resources = resources
+        s
+      }
+      
+      it 'calls check_ci_status' do
+        expect(subject).to receive(:check_ci_status)
+        subject.send(:validate_commit)
+      end
+
+      context 'when skip_ci_status is true' do
+        let(:skip_ci) { true }
+        it 'does not call check_ci_status' do
+          expect(subject).not_to receive(:check_ci_status)           
+          subject.send(:validate_commit)
+        end
+      end
+    end
+
   end
 end
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -67,6 +67,7 @@ new command line option:
 ```
 Usage: moonshot build VERSION
     -v, --[no-]verbose               Show debug logging
+    -s, --skip-ci-status             Skip checks on CI jobs
     -n, --environment=NAME           Which environment to operate on.
         --[no-]interactive-logger    Enable or disable fancy logging
     -F, --foo                        ENABLE HYPERDRIVE


docs/mechanisms/build.md
```diff
@@ -33,6 +33,18 @@ Moonshot.config do |c|
 ...
 ```
 
+Also **skip_ci_status** is an optional flag would allow us to skip checks
```
kaushik: This can be '**skip_ci_status** is an optional flag. It would allow us to skip checks'

lib/moonshot/build_mechanism/github_release.rb
```diff
@@ -159,7 +163,9 @@ def validate_commit
       sh_step(cmd, msg: "Validate commit #{@sha}.") do |_, out|
         @commit_detail = out
       end
-      @ci_statuses = check_ci_status(@sha)
+      @ci_statuses = check_ci_status(@sha) unless
```
pdrakeweb: I think you want to do this `if` the value is false, not `unless` the value is false, right?

lib/moonshot/build_mechanism/github_release.rb
```diff
@@ -17,10 +17,14 @@ class GithubRelease # rubocop:disable Metrics/ClassLength
...
     # @param build_mechanism Delegates building after GitHub release is created.
-    def initialize(build_mechanism, ci_status_timeout: 600, max_tag_find_timeout: 240)
+    def initialize(build_mechanism,
+                   ci_status_timeout: 600,
+                   max_tag_find_timeout: 240,
+                   skip_ci_status: false)
       @build_mechanism = build_mechanism
       @ci_status_timeout = ci_status_timeout
       @max_tag_find_timeout = max_tag_find_timeout
+      @skip_ci_status = skip_ci_status
```
pdrakeweb: Instead of supporting both this and the moonshot config, could we set this to the value passed (if one is) or default to the moonshot config value? `@skip_ci_status = skip_ci_status.nil? ? Moonshot.config.skip_ci_status : skip_ci_status` or something to that effect.

lib/moonshot/command.rb
```diff
@@ -15,13 +15,17 @@ def self.inherited(base)
 
     def parser
       @use_interactive_logger = true
-
+
```
pdrakeweb: Got some extra whitespace in there somehow.

lib/moonshot/command.rb
```diff
@@ -40,7 +44,6 @@ def controller
       config = Moonshot.config
       config.update_for_account!
       controller = Moonshot::Controller.new(config)
-
```
pdrakeweb: I don't think this whitespace should be deleted.

lib/moonshot/command.rb
```diff
@@ -23,6 +23,10 @@ def parser
           Moonshot.config.interactive_logger = InteractiveLogger.new(debug: true) if v
         end
 
+        o.on('-s', '--skip-ci-status', 'Skip checks on CI jobs') do
```
askreet: We added the ability for plugins and mechanisms to hook into the CLI parser when dispatching a command, we can use that here to avoid having this option exposed to classes that aren't going to use it.

That might look something like (assuming this is used during deploy):

In github_release.rb:
```ruby
def initialize(skip_ci_status: false)
  @skip_ci_status = skip_ci_status
end
```

```ruby
def deploy_cli_hook(parser)
  parser.on('-s', '--skip-ci-status', 'Skips checks on CI jobs', FalseClass) do |value|
    @skip_ci_status = value
  end
end
```

In this way, you completely own the configuration of the mechanism within itself.

lib/moonshot/build_mechanism/github_release.rb
```diff
@@ -17,10 +17,14 @@ class GithubRelease # rubocop:disable Metrics/ClassLength
...
     # @param build_mechanism Delegates building after GitHub release is created.
-    def initialize(build_mechanism, ci_status_timeout: 600, max_tag_find_timeout: 240)
+    def initialize(build_mechanism,
+                   ci_status_timeout: 600,
+                   max_tag_find_timeout: 240,
+                   skip_ci_status: false)
       @build_mechanism = build_mechanism
       @ci_status_timeout = ci_status_timeout
       @max_tag_find_timeout = max_tag_find_timeout
+      @skip_ci_status = skip_ci_status.nil? ? Moonshot.config.skip_ci_status : skip_ci_status
```
askreet: We default to false, which means that by default we won't use the Moonshot.config.skip_ci_status unless someone passes nil as a keyword argument here. If we implement CLI hooks on the build mechanisms, we can modify the instance state using the command line argument lambda of OptionParser, and allow a keyword argument to override the default.

docs/mechanisms/build.md
```diff
@@ -33,6 +33,18 @@ Moonshot.config do |c|
 ...
 ```
 
+**skip_ci_status** is an optional flag. It would allow us to skip checks 
+on the commit's CI job status. If its not finished yet, no need for waiting till timeout.
```
askreet: How about instead of "If...", we say "Without this option, the GithubRelease mechanism will wait until the build is finished."

askreet: Do you want to mention that a command-line option is available to override this value?

docs/user-guide/cli.md
```diff
@@ -18,6 +18,13 @@ custom plugins.
 
 #### `--environment=NAME / -n NAME`
 
+*Default: Disabled*
```
askreet: I think you've got this backwards, this appears to add documentation to the --environment= option.

lib/moonshot/build_mechanism/github_release.rb
```diff
@@ -17,10 +17,14 @@ class GithubRelease # rubocop:disable Metrics/ClassLength
     def_delegator :@build_mechanism, :output_file
 
     # @param build_mechanism Delegates building after GitHub release is created.
-    def initialize(build_mechanism, ci_status_timeout: 600, max_tag_find_timeout: 240)
+    def initialize(build_mechanism,
+                   ci_status_timeout: 600,
+                   max_tag_find_timeout: 240,
+                   skip_ci_status: false)
```
pdrakeweb: This should just be nil now, right?

lib/moonshot/build_mechanism/github_release.rb
hmagdy: Default argument value should be nil..  now Fixed

lib/moonshot/command.rb
hmagdy: I was start this feature with this approach, actually I like it to mush because I lean toward keeping every module as a standalone and ready for plug and play.

Then I moved to current approach based on 2 reasons:

1.    I would like to see that flag on --help command e.g.
```
moonshot build --help
Usage: moonshot build VERSION
    -v, --[no-]verbose               Show debug logging
    -s, --skip-ci-status             Skip checks on CI jobs
    -n, --environment=NAME           Which environment to operate on.
        --[no-]interactive-logger    Enable or disable fancy logging
```
2.    I thought all plugins,  added by Acquia team are part of command.rb as a backbone of Moonshot.
So now what you prefer?

pdrakeweb: Perhaps moonshot could be enhanced to call the proper hooks when command help is requested.

askreet: I agree with @pdrakeweb, but we can do that as a follow-up. We have other flags that are missing for core features because of the same reason, like CodeDeploy.

hmagdy: Could you take a look again @askreet @pdrakeweb also I think we have to add a story in JIra or a note to be done later for enhancing moonshot command line when --help flag is requested.

spec/moonshot/build_mechanism/github_release_spec.rb
```diff
@@ -189,5 +189,32 @@ module Moonshot # rubocop:disable ModuleLength
...
+      end
+
+      before(:each) do
+        allow(resources.ilog).to receive(:start_threaded).and_yield(step)
+        Retriable.configure do |c|
+          c.sleep_disabled = true
+        end
+      end
+
+      it 'passes if the validate commit return nil with skip_ci_status flag' do
```
pdrakeweb: I feel like it would be better to explicitly assert that check_ci_status is not run when the flag is set.  But I don't feel strongly.

askreet: Yeah, I think that's a good test, and it keeps the unit boundaries. In the current implementation the details of check_ci_status matter, and they shouldn't.

hmagdy: Fixed

lib/moonshot/command.rb
hmagdy: No need for --help hock .. I think now its working once plugin registered .. PR now updated with steps for checking --help flag and we can see
```
moonshot deploy --help 
Usage: moonshot deploy VERSION
    -v, --[no-]verbose               Show debug logging
    -n, --environment=NAME           Which environment to operate on.
        --[no-]interactive-logger    Enable or disable fancy logging
        --[no-]interactive           Use interactive prompts.
    -s, --skip-ci-status             Skips checks on CI jobs
        --ignore-app-stop-failures   Continue deployment on ApplicationStop failures
```

spec/moonshot/build_mechanism/github_release_spec.rb
```diff
@@ -189,5 +189,27 @@ module Moonshot # rubocop:disable ModuleLength
...
       end
     end
+    
+    describe '#validate_commit' do
+      let(:subject_skip_ci) {
+        s = described_class.new(build_mechanism , skip_ci_status: true)
+        s.resources = resources
+        s
+      }
+      it 'passes if the check_ci_status not callable with skip_ci_status ture' do
```
pdrakeweb: *true

spec/moonshot/build_mechanism/github_release_spec.rb
```diff
@@ -189,5 +189,27 @@ module Moonshot # rubocop:disable ModuleLength
         expect { subject.send(:check_ci_status, sha) }.to raise_error(Shell::CommandError)
       end
     end
+    
+    describe '#validate_commit' do
+      let(:subject_skip_ci) {
```
pdrakeweb: A better format for this would be to define (let) skip_ci as false (default) and have the subject use that value.  Place the example for that directly below.  Then create a context block below that and override skip_ci to be true and place the example for skip_ci being true in that context.  Pseudo-code:
```
describe '#validate_commit' do
let(:skip_ci) { false }
subject { described_class.new(build_mechanism, skip_ci_status: skip_ci); ... }

it 'calls check_ci_status' do
  ...
end

context 'when skip_ci_status is true' do
  let(:skip_ci) { true }

  it 'does not call check_ci_status' do
    ...
  end
end
```

hmagdy: Looks better and more readable for me 👍  now Done

lib/moonshot/command.rb
hmagdy: @askreet Could you take a look again,, also PR description updated

lib/moonshot/build_mechanism/github_release.rb
pdrakeweb: Resolved, no longer needs to be nil.

lib/moonshot/build_mechanism/github_release.rb
pdrakeweb: Resolved, this is no longer required.

PR #235, opened 2017-08-31
CPD-1298 | As a cloud data engineer, I would like to able to skip the SHA status check when building a release.

hmagdy:
```
moonshot build --help 
Usage: moonshot build VERSION
    -v, --[no-]verbose               Show debug logging
    -n, --environment=NAME           Which environment to operate on.
        --[no-]interactive-logger    Enable or disable fancy logging
    -s, --[no-]skip-ci-status        Skips checks on CI jobs
```

Testing command:
`rspec   spec/moonshot/commands/build_spec.rb`

this PR is tested and working perfect

```
[ ✓ ] [ 0m 22s ] Clone git@github.com:hmagdy/cloud-database-api.git to release-cloud-database-api-1.2.320170831-23703-i8ucvg                 
[ ✓ ] [ 0m 1s ] Checkout 4f43bfbe7edbe6cb950326ad60571a4d85e504f8                                                                            
[ ✓ ] [ 0m 1s ] rsync -a --include='vendor/bundle/***' --include='vendor*/' --exclude='*' /usr/src/aq/repos/cloud-database-api/ ./           
[ ✓ ] [ 24m 29s ] bundle install --deployment -j2                                                                                            
----> bundle exec moonshot build 1.2.3 --skip-ci-status
Skip CI Status flag in build_cli_hook is:
true
I, [2017-08-31T13:29:09.337836 #25897]  INFO -- : git fetch --tags upstream
I, [2017-08-31T13:29:15.051786 #25897]  INFO -- : Success
I, [2017-08-31T13:29:15.072169 #25897]  INFO -- : Validate commit 4f43bfbe7edbe6cb950326ad60571a4d85e504f8.
I, [2017-08-31T13:29:15.109533 #25897]  INFO -- : Success
Ci_statusesin after call check_ci_status is:
I, [2017-08-31T13:29:15.109716 #25897]  INFO -- : Validate `CHANGELOG.md`.
```

### DIFF
--- a/lib/moonshot/build_mechanism/github_release.rb
+++ b/lib/moonshot/build_mechanism/github_release.rb
@@ -27,14 +27,8 @@ module Moonshot::BuildMechanism
       @skip_ci_status = skip_ci_status
     end
 
-    def deploy_cli_hook(parser)
-      parser.on('-s', '--skip-ci-status', 'Skips checks on CI jobs', FalseClass) do |value|
-        @skip_ci_status = value
-      end
-    end
-
     def build_cli_hook(parser)
-      parser.on('-s', '--skip-ci-status', 'Skips checks on CI jobs', FalseClass) do |value|
+      parser.on('-s', '--[no-]skip-ci-status', 'Skips checks on CI jobs', TrueClass) do |value|
         @skip_ci_status = value
       end
     end
--- a/spec/moonshot/commands/build_spec.rb
+++ b/spec/moonshot/commands/build_spec.rb
@@ -1,0 +1,19 @@
+describe Moonshot::Commands::Build do
+
+  it 'should not handle --skip-ci-status and build correctly' do
+    cli_dispatcher = Moonshot::CommandLineDispatcher.new('build', subject, {})
+    parser = cli_dispatcher.send(:build_parser, subject)
+    expect { parser.parse(%w(--skip-ci-status)) }.to raise_error(RuntimeError)
+  end
+  
+  it 'should handle --skip-ci-status and build correctly' do
+    Moonshot.config = Moonshot::ControllerConfig.new
+    Moonshot.config do |c|
+      c.build_mechanism = Moonshot::BuildMechanism::GithubRelease.new('')
+    end
+    cli_dispatcher = Moonshot::CommandLineDispatcher.new('build', subject, {})
+    parser = cli_dispatcher.send(:build_parser, subject)
+    expect { parser.parse(%w(--skip-ci-status)) }.to_not raise_error
+  end 
+  
+end


kaushik:
Whats HA in the title?

lib/moonshot/build_mechanism/github_release.rb
```diff
@@ -28,14 +28,14 @@ def initialize(build_mechanism,
     end
 
     def deploy_cli_hook(parser)
```
kaushik: I still do not understand why we need a deploy hook for this, we AFAIK do no look for ci status during deploy. Correct me if I am wrong.

lib/moonshot/build_mechanism/github_release.rb
```diff
@@ -204,6 +204,7 @@ def fetch_changes(version)
     #
     # @return [String] Status and links to the CI jobs
     def check_ci_status(sha)
+      say("Checks for the commit's CI job status.", :yellow)
```
kaushik: I don't think this is required, we already say 'validating commit'

hmagdy: this for double check and making sure this function not callable if flag sent

pdrakeweb: I think we should probably drop this as its not required.

kaushik: Okay, May be change it ot 'Checking' since this is output runtime.

hmagdy: Fixed

lib/moonshot/build_mechanism/github_release.rb
hmagdy: deploy hock removed

spec/moonshot/commands/build_spec.rb
```diff
@@ -0,0 +1,19 @@
+describe Moonshot::Commands::Build do
```
kaushik: This test looks fine, I was thinking if its possible to put the skip-ci-status test in the test file covering github release if there is one.

hmagdy: there is another unit test for github release already merged in previous PR .. this is for build command calls ..  you can find it here
https://github.com/acquia/moonshot/blob/master/spec/moonshot/build_mechanism/github_release_spec.rb#L193

kaushik: okay

lib/moonshot/build_mechanism/github_release.rb
```diff
@@ -27,15 +27,9 @@ def initialize(build_mechanism,
...
-      parser.on('-s', '--skip-ci-status', 'Skips checks on CI jobs', FalseClass) do |value|
-        @skip_ci_status = value
-      end
-    end
-
     def build_cli_hook(parser)
-      parser.on('-s', '--skip-ci-status', 'Skips checks on CI jobs', FalseClass) do |value|
-        @skip_ci_status = value
+      parser.on('-s', '--skip-ci-status', 'Skips checks on CI jobs') do |value|
+        @skip_ci_status = true if value
```
borsothy: If I understand correctly, @askreet suggested this:
`parser.on('-s', '--skip-ci-status', 'Skips checks on CI jobs', FalseClass)`

I suppose the last parameter defines a default value, therefore it can be simplified to something like this:
`@skip_ci_status = !!value`
or even perhaps just
`@skip_ci_status = value`

Same applies for the other hook defined in the previous PR.
@askreet can you verify this please?

spec/moonshot/commands/build_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  it 'should not handle --skip-ci-status and build correctly' do
+    soso = Moonshot::CommandLineDispatcher.new('build', subject, {})
```
borsothy: Can you please change `soso` to something more self explanatory?

spec/moonshot/commands/build_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+    op = soso.send(:build_parser, subject)
```
borsothy: Same for `op`.

lib/moonshot/build_mechanism/github_release.rb
```diff
@@ -27,15 +27,9 @@ def initialize(build_mechanism,
...
-    def deploy_cli_hook(parser)
-      parser.on('-s', '--skip-ci-status', 'Skips checks on CI jobs', FalseClass) do |value|
-        @skip_ci_status = value
-      end
-    end
-
     def build_cli_hook(parser)
-      parser.on('-s', '--skip-ci-status', 'Skips checks on CI jobs', FalseClass) do |value|
-        @skip_ci_status = value
+      parser.on('-s', '--skip-ci-status', 'Skips checks on CI jobs') do |value|
```
borsothy: I wonder if `-s` is verbose enough to be used in this case, but I'm not sure if we can get rid of that 🤔

lib/moonshot/build_mechanism/github_release.rb
hmagdy: No need for default value or even assign the received value.. because of no value sent in command parameter .. so if this flag sent that means true else there is default value in constructor

spec/moonshot/commands/build_spec.rb
hmagdy: Fixed

spec/moonshot/commands/build_spec.rb
hmagdy: Fixed too ,, P.S. this variable name used in all commands spec tests .. may be refactoring should be applied later

lib/moonshot/build_mechanism/github_release.rb
hmagdy: Fixed

spec/moonshot/commands/build_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+    parser = cli_dispatcher.send(:build_parser, subject)
+    expect { parser.parse(%w(--skip-ci-status)) }.to raise_error(RuntimeError)
+  end
+  
+  it 'should handle --skip-ci-status and build correctly' do
+    Moonshot.config = Moonshot::ControllerConfig.new
+    Moonshot.config do |c|
+      c.build_mechanism = Moonshot::BuildMechanism::GithubRelease.new('')
+    end
+    cli_dispatcher = Moonshot::CommandLineDispatcher.new('build', subject, {})
```
glennpratt: The difference between these two test cases is not clear from the description.

The one with non-standard config should have a clearer description or use a context to do that.

lib/moonshot/build_mechanism/github_release.rb
```diff
@@ -27,14 +27,8 @@ def initialize(build_mechanism,
...
 
-    def deploy_cli_hook(parser)
-      parser.on('-s', '--skip-ci-status', 'Skips checks on CI jobs', FalseClass) do |value|
-        @skip_ci_status = value
-      end
-    end
-
     def build_cli_hook(parser)
-      parser.on('-s', '--skip-ci-status', 'Skips checks on CI jobs', FalseClass) do |value|
+      parser.on('-s', '--[no-]skip-ci-status', 'Skips checks on CI jobs', TrueClass) do |value|
```
glennpratt: huh, TrueClass is a weird argument to take. I'll assume this all makes sense though :)

PR #237, opened 2017-09-15
Added ability to push CF template to S3

smatyas:
I hit the [51,200 bytes size limitation](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cloudformation-limits.html) of the template body.
One of the suggested resolutions there is to upload the template to an S3 bucket. This patch makes moonshot to be able to do that.

I added a new configuration option: `template_s3_bucket`. If it's set, moonshot will always upload the local template to that bucket then use its URL in the API calls.


smatyas:
closing in favor of https://github.com/acquia/moonshot/pull/239

PR #239, opened 2017-11-06
Added ability to push CF template to S3

smatyas:
I hit the [51,200 bytes size limitation](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cloudformation-limits.html) of the template body.
One of the suggested resolutions there is to upload the template to an S3 bucket. This patch makes moonshot to be able to do that.

I added a new configuration option: `template_s3_bucket`. If it's set, moonshot will always upload the local template to that bucket then use its URL in the API calls.

### DIFF
--- a/lib/moonshot/stack.rb
+++ b/lib/moonshot/stack.rb
@@ -177,15 +177,15 @@ module Moonshot
     end
 
     def upload_template_to_s3
-      raise 'The S3 bucket to store the template in is not configured.' unless @config.template_s3_bucket
+      raise 'The S3 bucket to store the template in is not configured.' unless @config.template_s3_bucket # rubocop:disable LineLength
 
       s3_object_key = "#{@name}-#{File.basename(template.filename)}"
-        s3_client = Aws::S3::Client.new
-        s3_client.put_object(
-          bucket: @config.template_s3_bucket,
-          key: s3_object_key,
-          body: template.body
-        )
+      s3_client = Aws::S3::Client.new
+      s3_client.put_object(
+        bucket: @config.template_s3_bucket,
+        key: s3_object_key,
+        body: template.body
+      )
       template_url = "http://#{@config.template_s3_bucket}.s3.amazonaws.com/#{s3_object_key}"
       puts "The template has been uploaded to #{template_url}"
 
--- a/lib/moonshot/stack.rb
+++ b/lib/moonshot/stack.rb
@@ -176,14 +176,35 @@ module Moonshot
       raise "Could not describe stack: #{name}"
     end
 
+    def upload_template_to_s3
+      raise 'The S3 bucket to store the template in is not configured.' unless @config.template_s3_bucket
+
+      s3_object_key = "#{@name}-#{File.basename(template.filename)}"
+        s3_client = Aws::S3::Client.new
+        s3_client.put_object(
+          bucket: @config.template_s3_bucket,
+          key: s3_object_key,
+          body: template.body
+        )
+      template_url = "http://#{@config.template_s3_bucket}.s3.amazonaws.com/#{s3_object_key}"
+      puts "The template has been uploaded to #{template_url}"
+
+      template_url
+    end
+
     def create_stack
-      cf_client.create_stack(
+      parameters = {
         stack_name: @name,
-        template_body: template.body,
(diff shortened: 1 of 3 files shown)


smatyas:
@askreet would you please consider this addition? The SFNG project depends on this, it would be great to have it in moonshot.

lib/moonshot/stack.rb
```diff
@@ -176,14 +176,35 @@ def get_stack(name)
       raise "Could not describe stack: #{name}"
     end
 
+    def upload_template_to_s3
+      raise 'The S3 bucket to store the template in is not configured.' unless @config.template_s3_bucket # rubocop:disable LineLength
```
askreet: Please reformat this to a multi-line conditional, rather than disabling rubocop.

lib/moonshot/stack.rb
```diff
@@ -176,14 +176,35 @@ def get_stack(name)
...
+
+      s3_object_key = "#{@name}-#{File.basename(template.filename)}"
```
askreet: For modern Moonshot users, this will always be "template", since the conventional location is [project root]/moonshot/template.yml. One thing I don't like about S3-based workflow is overwriting files loses you the context of historical stack state. Perhaps we should relate the filename to a timestamp or UUID, then we can see historical changes.

askreet: See [here](https://github.com/acquia/moonshot/blob/master/lib/moonshot/stack.rb#L153).

lib/moonshot/stack.rb
```diff
@@ -176,14 +176,35 @@ def get_stack(name)
...
+      s3_client = Aws::S3::Client.new
+      s3_client.put_object(
+        bucket: @config.template_s3_bucket,
+        key: s3_object_key,
+        body: template.body
+      )
+      template_url = "http://#{@config.template_s3_bucket}.s3.amazonaws.com/#{s3_object_key}"
+      puts "The template has been uploaded to #{template_url}"
```
askreet: For things that happen during create/update/delete lifecycle, please use @ilog's threaded logging. Take a look at  `#delete` to get an idea of what to do.

spec/moonshot/stack_spec.rb
```diff
@@ -88,10 +90,32 @@
...
+            bucket: config.template_s3_bucket,
+            key: an_instance_of(String),
+            body: an_instance_of(String)
+          }
+        end
+
+        it 'should call put_object' do
+          expect(s3_client).to receive(:put_object)
+            .with(hash_including(expected_put_object_options))
+          expect(cf_client).to receive(:create_stack)
```
askreet: I think we also want to make an assertion about the create_stack call, because it should have `template_url` in place of `template_body`.

spec/moonshot/stack_spec.rb
```diff
@@ -88,10 +90,32 @@
...
+
+      context 'when template_s3_bucket is set' do
+        before(:each) do
+          config.template_s3_bucket = 'rspec-bucket'
+        end
+
+        let(:expected_put_object_options) do
+          {
+            bucket: config.template_s3_bucket,
+            key: an_instance_of(String),
```
askreet: If we mock out some of the components used in the key, we can assert the behavior here instead of just expecting an instance of a string (e.g. Time).

lib/moonshot/stack.rb
```diff
@@ -177,17 +177,22 @@ def get_stack(name)
...
+      unless @config.template_s3_bucket
+        raise 'The S3 bucket to store the template in is not configured.'
+      end
+
+      s3_object_key = "#{@name}-#{Time.now.getutc.to_i}-#{File.basename(template.filename)}"
       template_url = "http://#{@config.template_s3_bucket}.s3.amazonaws.com/#{s3_object_key}"
-      puts "The template has been uploaded to #{template_url}"
+
+      @ilog.start "Uploading template to #{template_url}" do |s|
+        s3_client = Aws::S3::Client.new
```
askreet: We already have a method for this in CredsHelper. I think you may just be able to remove this line.

smatyas:
Thanks for the suggestions @askreet I think I addressed everything. Could take another look on this, please?

smatyas:
@janelletavares please take it over from me, we can switch back our dependency once it gets released.

PR #256, opened 2019-07-30
CPD-4445: Fix dynamic template plugin issue with reading cli parameters.

valentyntymku:
CPD-4445: Fix dynamic template plugin issue with reading cli parameters.

### DIFF
--- a/docs/plugins/dynamic_template.md
+++ b/docs/plugins/dynamic_template.md
@@ -10,7 +10,7 @@ available in the form of a CLI.
 
 The plugin accepts three configuration parameters:
 - `source`: path to the source template file
-- `parameters`: key-value pairs of the parameters of your choice
+- `parameters`: parameters of your choice (see below - [parameters](#parameters))
 - `destination`: the target file path
 
 Your template must be conforming the ERB standards. You can use both
@@ -24,3 +24,22 @@ Moonshot::Plugins::DynamicTemplate.new(
   destination: 'cloud_formation/cdb-api.json'
 )
 ```
+
+## Parameters
+
+The 'hack' with lambdas created for the cases when you need to read
+CLI parameters before running the plugin.
+
+Parameters could be described in two ways:
+* key-value pairs   
+  ```ruby
+  parameters: { deletion_policy: 'Retain' }
+  ```
+
+* lambdas
+  ```ruby
+  parameters = -> environment_name {
+  environment = environment_name =~ prod_regexp ? 'prod' : 'dev'
+  parameters = YAML.load_file("cloud_formation/parameters/#{environment}.yml")
+  }
+  ```
--- a/lib/moonshot/dynamic_template.rb
+++ b/lib/moonshot/dynamic_template.rb
@@ -24,8 +24,17 @@ module Moonshot
 
     def initialize(source:, parameters:, destination:)
       @source = File.read(source)
-      @parameters = Parameters.new(parameters)
+      @parameters = parameters
       @destination = destination
+    end
+
+    def parameters_obj
+      @parameters_obj ||= Parameters.new(parameters_file)
+    end
+
+    def parameters_file
+      env_name = Moonshot.config.environment_name
+      @parameters.respond_to?(:call) ? @parameters.call(env_name) : @parameters
     end
 
     def process
@@ -52,7 +61,7 @@ module Moonshot
     end
 
     def generate_template
-      ERB.new(@source).result(@parameters.expose_binding)
+      ERB.new(@source).result(parameters_obj.expose_binding)
     end
 
     def write_output(content)
--- a/spec/moonshot/dynamic_template_spec.rb
+++ b/spec/moonshot/dynamic_template_spec.rb
@@ -4,17 +4,26 @@ describe Moonshot::DynamicTemplate do
   include FakeFS::SpecHelpers
 
   let(:source_path) { 'spec_template.erb' }
-  let(:parameters) { { test_param: 'test-value' } }
+  let(:parameters_lambda) { -> environment_name { parameters = { test_param: 'test-value' } } }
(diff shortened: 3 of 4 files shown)


lib/moonshot/dynamic_template.rb
```diff
@@ -24,12 +24,15 @@ def expose_binding
 
     def initialize(source:, parameters:, destination:)
       @source = File.read(source)
-      @parameters = Parameters.new(parameters)
+      @parameters = parameters
```
pdrakeweb: We should stop accepting parameters in `initialize` if we are going to overwrite it during `process`.

spec/moonshot/dynamic_template_spec.rb
```diff
@@ -4,7 +4,7 @@
   include FakeFS::SpecHelpers
```
pdrakeweb: I think we need a regression test that fails with the old code and passes with the new code.  I'm not sure if this change to the parameters `let` accomplishes that.  Does it?

lib/moonshot/dynamic_template.rb
```diff
@@ -24,12 +24,15 @@ def expose_binding
...
       @destination = destination
     end
 
     def process
       validate_destination_exists
+      @parameters = Parameters.new(
```
malikbakt: please change it in backwards compatible way (accepting both string and lambda)

```ruby

def parameters_file
  @parameters.respond_to?(:call) ? @parameters.call(Moonshot.config.environment_name) : @parameters
end

```

valentyntymku: Improvement done.
Now it accepts both: lambda and array of params.
Thanks.

spec/moonshot/dynamic_template_spec.rb
valentyntymku: Now it accepts both ways: lambda and string. So both code old and new are accepted.

lib/moonshot/dynamic_template.rb
valentyntymku: I guess we should accept parameter during initialization because in this way we are sending either lambda or array of params to the plugin.

spec/moonshot/dynamic_template_spec.rb
```diff
@@ -4,7 +4,7 @@
...
 
   let(:source_path) { 'spec_template.erb' }
-  let(:parameters) { { test_param: 'test-value' } }
+  let(:parameters) { -> environment_name { parameters = { test_param: 'test-value' } } }
```
malikbakt: please add specs for the case when parameters is a Hash.

lib/moonshot/dynamic_template.rb
borsothy: @pdrakeweb we should, @valentyntymku please provide an example how the plugin should be configured in the Moonfile. You might consider this to be slightly hacky, however this is a solution that doesn't require major modifications in Moonshot.

valentyntymku: We can use this in both ways in Moonshot.rb.
The 'old-school':
`parameters = { deletion_policy: 'Delete' }`

And the Lambda way:
`parameters = -> environment_name {
environment = environment_name =~ prod_regexp ? 'prod' : 'dev'
parameters = YAML.load_file("cloud_formation/parameters/#{environment}.yml")
}`

spec/moonshot/dynamic_template_spec.rb
pdrakeweb: Ok, we need another test then.  We should have a regression test that would have caught the original bug.

valentyntymku:
> @valentyntymku please update the documentation.

Docs updated.

spec/moonshot/dynamic_template_spec.rb
borsothy: @pdrakeweb the bug didn't really lie in the plugin itself, but rather in the way we _configured_ it, the incorrect set of values have been passed on. That is a "bug" in the `ControllerConfig` class which we chose not to address (it would have been extremely complicated to tackle the issue), but created a workaround instead. Long story short, I think a regression test is not possible in this case.

spec/moonshot/dynamic_template_spec.rb
valentyntymku: Done. Thanks.

docs/plugins/dynamic_template.md
```diff
@@ -10,7 +10,7 @@ available in the form of a CLI.
 
 The plugin accepts three configuration parameters:
 - `source`: path to the source template file
-- `parameters`: key-value pairs of the parameters of your choice
+- `parameters`: parameters of your choice (see below)
```
marciotoshio: it could use a anchor link

valentyntymku: Created anchor link. Thanks.

spec/moonshot/dynamic_template_spec.rb
```diff
@@ -31,30 +40,82 @@
...
+  end
+
+  describe '#validate_template' do
+    before(:each) do
+      allow(cf_client).to receive(:validate_template).and_raise(
+        Aws::CloudFormation::Errors::ValidationError.new(nil, nil)
+      )
+    end
+
+    it 'should raise an exception if the tempalte is invalid' do
```
pdrakeweb: *template

valentyntymku: Fixed. Thanks.

valentyntymku:
> This still seems to be missing a regression test. Ideally, we would have a test that can be run on the old code and exposes the escaped defect. And which passes on the new code. The added unit tests are good, but none appear to be a regression test or prove that the escaped defect can't reoccur.

If we take into account Gergely's comment than a regression test is not possible in this case. ([Comment](https://github.com/acquia/moonshot/pull/256#discussion_r311500623))

PR #257, opened 2019-08-07
CPD-3751: Enabling interactive option on the build command.

marciotoshio:


lib/moonshot/commands/interactive_command.rb
```diff
@@ -0,0 +1,13 @@
+module Moonshot
+  module Commands
+    class InteractiveCommand < Moonshot::Command
```
jamesiarmes: I feel like this should be a module rather than a class. This would allow commands to import various features, though we may only have interactive for now, without having to pick one to extend from or implementing an ever increasingly complex set of base classes.

PR #267, opened 2020-09-21
CPD-183: Added ASG rotation plugin.

AbhishekRana95:
Change: minor
Purpose: feature

### Manual review steps:

1. For pointing the changes for local environment of moonshot gem: [How to specify local Ruby gems in your Gemfile](https://rossta.net/blog/how-to-specify-local-ruby-gems-in-your-gemfile.html)
2. Once the above step is up and working, include the plugin in cloud-database-worker repo in Moonfile.rb by adding `c.plugins << Moonshot::Plugins::RotateAsgInstances.new` to the file.
3. Update the AMIMap key in `cdb-worker.json` and change the AMIs.
4. Run `cd cloud-database-worker && bundle exec moonshot update`
5. Open the aws console and go to `CloudFormation > Stacks > cdb-worker-dev-#{stack_name}` where stack_name is your worker stack.
6. Go to the ASG of the stack and check the old instances are removed and new instances are attached with the updated launch configuration for ASG.

> We should also verify that the outdated instances currently in the ASG are stopped correctly.

### DIFF
--- a/docs/plugins/rotate_asg_instances.md
+++ b/docs/plugins/rotate_asg_instances.md
@@ -1,0 +1,26 @@
+# Rotate ASG Instances Plugin
+
+## Overview
+The rotate ASG instances plugin rotates the outdated instances in Auto Scaling Groups. It compares the launch configuration and sees if any outdated instances are present. It detaches the instances first then shuts down the instance and waits for a new instance to replace the outdated one in ASG then proceeds to the next outdated instance.
+
+After all outdated instances are shutdown successfully, it terminates them and reaps the associated volumes.
+
+## Usage
+It allows gracefully shutting down each instance instead of terminating them and killing all the running processes.
+
+## Configuration
+The plugin uses the ssh username specified in the `MOONSHOT_SSH_USER` or the `LOGNAME` environment variable for logging into the ASG instances to shutdown. The value should be the username with which you have the access to the instances. For example:
+```ruby
+export MOONSHOT_SSH_USER=abhishek.rana
+```
+
+The plugin needs no additional configuration parameters:
+
+## Example
+```ruby
+Moonshot.config do |c|
+  # ...
+  c.plugins << Moonshot::Plugins::RotateAsgInstances.new
+  # ...
+end
+```
--- a/lib/plugins/rotate_asg_instances.rb
+++ b/lib/plugins/rotate_asg_instances.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+require 'aws-sdk'
+require_relative 'rotate_asg_instances/asg'
+
+module Moonshot
+  module Plugins
+    # Rotate ASG instances after update.
+    class RotateAsgInstances
+      def post_update(resources)
+        asg = ASG.new(resources)
+        asg.rotate_asg_instances
+        asg.teardown_outdated_instances
+      end
+    end
+  end
+end
--- a/lib/plugins/rotate_asg_instances/asg.rb
+++ b/lib/plugins/rotate_asg_instances/asg.rb
@@ -1,0 +1,238 @@
+require 'moonshot/ssh_fork_executor'
+
+module Moonshot
+  class ASG # rubocop:disable Metrics/ClassLength
+    include Moonshot::CredsHelper
+
+    def initialize(resources)
+      @resources = resources
+      @ilog = @resources.ilog
+      @ssh_user = ENV['MOONSHOT_SSH_USER'] || ENV['LOGNAME']
+    end
+
+    def asg
+      @asg ||=
+        Aws::AutoScaling::AutoScalingGroup.new(name: physical_resource_id)
+    end
+
+    def rotate_asg_instances
+      @ilog.start_threaded('Rotating ASG instances...') do |step|
+        @step = step
+        @volumes_to_delete = outdated_volumes(outdated_instances)
+        @shutdown_instances = cycle_instances(outdated_instances)
+        @step.success('ASG instances rotated successfully!')
+      end
+    end
+
+    def teardown_outdated_instances
+      @ilog.start_threaded('Tearing down outdated instances...') do |step|
+        @step = step
+        terminate_instances(@shutdown_instances)
+        reap_volumes(@volumes_to_delete)
+        @step.success('Outdated instances removed successfully!')
+      end
+    end
+
+    def physical_resource_id
+      @resources.controller.stack
+                .resources_of_type('AWS::AutoScaling::AutoScalingGroup')
+                .first.physical_resource_id
+    end
+
+    def outdated_instances
+      @outdated_instances ||=
+        asg.instances.reject do |i|
+          i.launch_configuration_name == asg.launch_configuration_name
+        end
+    end
+
+    private
+
+    def outdated_volumes(outdated_instances)
+      volumes = []
+      outdated_instances.each do |i|
+        begin
+          inst = Aws::EC2::Instance.new(id: i.id)
+          volumes << inst.block_device_mappings.first.ebs.volume_id
+        rescue StandardError => e
+          # We're catching all errors here, because failing to reap a volume
+          # is not a critical error, will not cause issues with the update.
+          @step.failure('Failed to get volumes for instance '\
+                    "#{i.instance_id}: #{e.message}")
+        end
+      end
+      volumes
+    end
+
+    # Cycle the instances in the ASG.
+    #
+    # Each instance will be detached one at a time, waiting for the new instance
+    # to be ready before stopping the worker and terminating the instance.
+    #
+    # @param instances [Array] (outdated instances)
+    #   List of instances to cycle. Defaults to all instances with outdated
+    #   launch configurations.
+    # @return [Array] (array of Aws::AutoScaling::Instance)
+    #   List of shutdown instances.
+    def cycle_instances(outdated_instances)
+      shutdown_instances = []
+
+      if outdated_instances.empty?
+        @step.success('No instances cycled!')
+        return []
+      end
+
+      @step.success("Cycling #{outdated_instances.size} " \
+                    "of #{asg.instances.size} instances in " \
+                    "#{physical_resource_id}...")
+
+      # Iterate over the instances in the stack, detaching and terminating each
+      # one.
+      outdated_instances.each do |i|
+        next if %w(Terminating Terminated).include?(i.lifecycle_state)
+
+        wait_for_instance(i)
+        detach_instance(i)
+
+        @step.success("Shutting down #{i.instance_id}")
+        shutdown_instance(i.instance_id)
+        shutdown_instances << i
+      end
+
+      @step.success('All instances cycled.')
+
+      shutdown_instances
+    end
+
+    # Waits for an instance to reach a ready state.
+    #
+    # @param instance [Aws::AutoScaling::Instance] Auto scaling instance to wait
+    #   for.
+    def wait_for_instance(instance, state = 'InService')
+      instance.wait_until(max_attempts: 60, delay: 10) do |i|
+        i.lifecycle_state == state
+      end
+    end
+
+    # Detach an instance from its ASG. Re-attach if failed.
+    #
+    # @param instance [Aws::AutoScaling::Instance] Instance to detach.
+    def detach_instance(instance)
+      @step.success("Detaching instance: #{instance.instance_id}")
+
+      # If the ASG can't be brought up to capacity, re-attach the instance.
+      begin
+        instance.detach(should_decrement_desired_capacity: false)
+        @step.success('- Waiting for the AutoScaling '\
+                     'Group to be up to capacity')
+        wait_for_capacity
+      rescue StandardError => e
+        @step.failure("Error bringing the ASG up to capacity: #{e.message}")
+        @step.failure("Attaching instance: #{instance.instance_id}")
+        reattach_instance(instance)
+        raise e
+      end
+    end
+
+    # Re-attach an instance to its ASG.
+    #
+    # @param instance [Aws::AutoScaling::Instance] Instance to re-attach.
+    def reattach_instance(instance)
+      instance.load
+      return unless instance.data.nil? \
+        || %w(Detached Detaching).include?(instance.lifecycle_state)
+
+      until instance.data.nil? || instance.lifecycle_state == 'Detached'
+        sleep 10
+        instance.load
+      end
+      instance.attach
+    end
+
+    # Terminate instances.
+    #
+    # @param instances [Array] (instances for termination)
+    #   List of instances to terminate. Defaults to all instances with outdated
+    #   launch configurations.
+    def terminate_instances(outdated_instances)
+      if outdated_instances.any?
+        @step.continue(
+          "Terminating #{outdated_instances.size} outdated instances..."
+        )
+      end
+      outdated_instances.each do |asg_instance|
+        instance = Aws::EC2::Instance.new(asg_instance.instance_id)
+        begin
+          instance.load
+        rescue Aws::EC2::Errors::InvalidInstanceIDNotFound
+          next
+        end
+
+        next unless %w(stopping stopped).include?(instance.state.name)
+
+        instance.wait_until_stopped
+
+        @step.continue("Terminating #{instance.instance_id}")
+        instance.terminate
+      end
+    end
+
+    def reap_volumes(volumes)
+      volumes.each do |volume_id|
+        begin
+          @step.continue("Deleting volume: #{volume_id}")
+          ec2_client(region: ENV['AWS_REGION'])
+            .delete_volume(volume_id: volume_id)
+        rescue StandardError => e
+          # We're catching all errors here, because failing to reap a volume
+          # is not a critical error, will not cause issues with the release.
+          @step.failure("Failed to delete volume #{volume_id}: #{e.message}")
+        end
+      end
+    end
+
+    # Waits for the ASG to reach the desired capacity.
+    def wait_for_capacity
+      @step.continue(
+        'Replacing outdated instances with new instances for the AutoScaling Group...'
+      )
+      # While we wait for the asg to reach capacity, report instance statuses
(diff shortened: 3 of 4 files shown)


docs/plugins/rotate_worker_instances.md
```diff
@@ -0,0 +1,19 @@
+# Dynamic template plugin
```
borsothy, line 1: I believe this was accidentally left unchanged after a copy-paste.

docs/plugins/rotate_worker_instances.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Overview
+The rotate worker instances plugins rotates the outdated instances in Auto Scaling Group. It compares the launch congiguration and see if any outdated instances are present. It detaches the instances first, then drain the workflows so that the instances should complete any pending jobs and not take any new tasks. After shutting down the instance it waits for new instance to replace the outdated one in ASG then proceed for the next outdated instance.
```
borsothy, line 4: Workflow draining is a feature of our AMI, this logic doesn't include that.

lib/plugins/rotate_worker_lib/asg.rb
```diff
@@ -0,0 +1,252 @@
+require_relative 'system'
+
+module Moonshot
+  class ASG # rubocop:disable Metrics/ClassLength
+    include Moonshot::CredsHelper
+
+    def initialize(resources)
+      @resources = resources
+      @ilog = @resources.ilog
+      @ssh_user = ENV['CDB_CI_USER'] || ENV['LOGNAME']
```
borsothy, line 10: "CDB_CI_USER" is specific to another service, we should remove that.

lib/plugins/rotate_worker_lib/asg.rb
```diff
@@ -0,0 +1,252 @@
...
+    def instances
+      asg.instances
+    end
+
+    # Retrieves the current ASG launch configuration name.
+    #
+    # @return [String]
+    def launch_configuration_name
+      asg.launch_configuration_name
+    end
```
borsothy, line 53: I don't feel strongly, but these wrapper methods look a bit unneccessary.

lib/plugins/rotate_worker_lib/asg.rb
```diff
@@ -0,0 +1,252 @@
...
+          i.lifecycle_state == 'InService'
+        end
+        instances_up.length == a.desired_capacity
+      end
+      @step.success('AutoScaling Group up to capacity!')
+    end
+
+    # Shuts down an instance, waiting for the worker to stop processing requests
+    # first. We do this instead of using the API so that
+    # services will be stopped properly.
```
borsothy, line 231: Phrases "API" and "worker" are specific to a service, we should remove these.

spec/moonshot/plugins/asg.rb
```diff
@@ -0,0 +1,285 @@
...
+      stub_cf_client
+    end
+    context 'when CDB_CI_USER is not defined' do
+      it 'uses LOGNAME for ssh_user' do
+        ENV['LOGNAME'] = 'SemiCoolDude'
+        ENV.delete('CDB_CI_USER')
+        expect(subject.instance_variable_get(:@ssh_user))
+          .to eq('SemiCoolDude')
+      end
+    end
```
borsothy, line 124: Please rename/remove everything related to "CDB_CI_USER".

borsothy (changes requested):

AbhishekRana95:
@borsothy I have addressed all your comments. Please let me know if anything else needs to be changed.

lib/plugins/rotate_worker_lib/asg.rb
borsothy: I just realized you use these terminologies in the name of the plugin too, let's remove that as well.

AbhishekRana95: Yes makes sense

docs/plugins/rotate_asg_instances.md
```diff
@@ -0,0 +1,24 @@
+# Rotate ASG Instances Plugin
+
+## Overview
+The rotate ASG instances plugin rotates the outdated instances in Auto Scaling Group. It compares the launch congiguration and see if any outdated instances are present. It detaches the instances first then shuts down the instance and waits for a new instance to replace the outdated one in ASG then proceed for the next outdated instance.
```
stephenthedev, line 4:
```suggestion
The rotate ASG instances plugin rotates the outdated instances in Auto Scaling Groups. It compares the launch configuration and see if any outdated instances are present. It detaches the instances first then shuts down the instance and waits for a new instance to replace the outdated one in ASG then proceeds to the next outdated instance.
```

docs/plugins/rotate_asg_instances.md
```diff
@@ -0,0 +1,24 @@
...
+
+After all outdated instances are shutdown successfully. It terminates the outdated instances and reap the volumes.
```
stephenthedev, line 6:
```suggestion
After all outdated instances are shutdown successfully, it terminates them and reaps the associated volumes.
```

docs/plugins/rotate_asg_instances.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Usage and configuration
+
+The plugin uses ENV['MOONSHOT_SSH_USER'] for logging into the ASG instances to shutdown. If ENV['MOONSHOT_SSH_USER'] is not set it uses ENV['LOGNAME'] for ssh purpose. The value should be the username with which you have the access to the instances. For example:
```
stephenthedev, line 10:
```suggestion
The plugin uses the ssh username specified in the `MOONSHOT_SSH_USER` or the `LOGNAME` environment variable for logging into the ASG instances to shutdown. The value should be the username with which you have the access to the instances. For example:
```

lib/plugins/rotate_asg_instances.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
```
stephenthedev, line 1: Do we need this?

lib/plugins/rotate_asg_instances.rb
```diff
@@ -0,0 +1,17 @@
...
+
+require 'aws-sdk'
+require_relative 'rotate_asg_instances_lib/asg'
+
+module Moonshot
+  module Plugins
+    # Rotate ASG instances after update.
```
stephenthedev, line 8: This comment seems to be already explained (yay readable code). Do we need it?

lib/plugins/rotate_asg_instances_lib/asg.rb
```diff
@@ -0,0 +1,239 @@
...
+        @step.success('ASG instances rotated successfully.')
+      end
+    end
+
+    def teardown_outdated_instances
+      @ilog.start_threaded('Tearing down outdated instances...') do |step|
+        @step = step
+        terminate_instances(@shutdown_instances)
+        reap_volumes(@volumes_to_delete)
+        @step.success('Outdated instances removed successfully!')
```
stephenthedev, line 31: Nit: This one uses ! at the end, and the other messages don't 😂 Let's be positive everywhere! lol

lib/plugins/rotate_asg_instances_lib/asg.rb
```diff
@@ -0,0 +1,239 @@
...
+      end
+    end
+
+    def name
+      @resources.controller.stack
+                .resources_of_type('AWS::AutoScaling::AutoScalingGroup')
+                .first.physical_resource_id
```
stephenthedev, line 38: Out of naivity, is this actually the name of the autoscaling group, or should we update this method name?

lib/plugins/rotate_asg_instances_lib/asg.rb
```diff
@@ -0,0 +1,239 @@
...
+
+    def outdated_instances
+      asg.instances.reject do |i|
+        i.launch_configuration_name == asg.launch_configuration_name
+      end
+    end
+
+    private
+
+    def outdated_volumes(instances = outdated_instances)
```
stephenthedev, line 49: Is there a reason we expose the option to pass a different set of instances other than just calling `outdated_instances` within the function block?

lib/plugins/rotate_asg_instances_lib/asg.rb
```diff
@@ -0,0 +1,239 @@
...
+    #
+    # Each instance will be detached one at a time, waiting for the new instance
+    # to be ready before stopping the worker and terminating the instance.
+    #
+    # @param instances [Array] (outdated instances)
+    #   List of instances to cycle. Defaults to all instances with outdated
+    #   launch configurations.
+    # @return [Array] (array of Aws::AutoScaling::Instance)
+    #   List of shutdown instances.
+    def cycle_instances(instances = outdated_instances)
```
stephenthedev, line 75: Same comment here about the method params :-)

lib/plugins/rotate_asg_instances_lib/system.rb
```diff
@@ -0,0 +1,80 @@
+class System
```
kaushik, line 1: Can you check if https://github.com/acquia/moonshot/blob/master/lib/moonshot/ssh_fork_executor.rb can be used instead of adding this new class?

kaushik (changes requested):
I was looking around and found https://github.com/acquia/moonshot/blob/ab44b072e82747eccddb59b6afc081b4ae5524d7/docs/plugins/asg_rollout.md. I could not dig deeper, but can you take a look that plugin and check if we are not duplicating the same behaviour or if we can extend it's behaviour.

lib/plugins/rotate_asg_instances.rb
pdrakeweb: This is community-accepted code style.  I think we should leave it.

lib/plugins/rotate_asg_instances.rb
pdrakeweb: This is community-accepted code style.  I think we should leave it.

lib/plugins/rotate_asg_instances_lib/asg.rb
pdrakeweb: Agreed.  We should just use `outdated_instances`.

pdrakeweb (commented):
Looks good.  I think this will be in good shape when the comments are addressed. 👍

lib/plugins/rotate_asg_instances_lib/asg.rb
AbhishekRana95: Agreed

AbhishekRana95:
> I was looking around and found https://github.com/acquia/moonshot/blob/ab44b072e82747eccddb59b6afc081b4ae5524d7/docs/plugins/asg_rollout.md. I could not dig deeper, but can you take a look that plugin and check if we are not duplicating the same behaviour or if we can extend it's behaviour.

This asg_rollout is introducing a new instance in ASG first and then removing the old instance. Let me see if I am able to reuse this and if not then find the reason why. Thank you for pointing this out @kaushik 💯

AbhishekRana95:
> > I was looking around and found https://github.com/acquia/moonshot/blob/ab44b072e82747eccddb59b6afc081b4ae5524d7/docs/plugins/asg_rollout.md. I could not dig deeper, but can you take a look that plugin and check if we are not duplicating the same behaviour or if we can extend it's behaviour.
> 
> This asg_rollout is introducing a new instance in ASG first and then removing the old instance. Let me see if I am able to reuse this and if not then find the reason why. Thank you for pointing this out @kaushik 💯

@kaushik there is a case in this that suppose there is no outdated instances in the ASG at the moment and we run `bundle exec moonshot update` in the repo. What happens is it first increase the desired capacity by 1, wait for and add an instance and then start the logic where it finds the outdated instances if not then what it does is that it decreases the capacity by 1 i.e. getting it back to normal. When the capacity is brought back to normal what AWS does is it terminates one of the instance in respond to the change and this termination is not doing it in the way we want and doing it according to the AWS termination logic which might lead to abruptly stopping the instance which might be executing some workflow at the time.

Let me know your thoughts on this 🤔

lib/plugins/rotate_asg_instances_lib/asg.rb
AbhishekRana95: We can change it to `physical_resource_it`

lib/plugins/rotate_asg_instances_lib/system.rb
AbhishekRana95: I think this class has way more to provide than the `ssh_fork_executor` and the implementation is also different in `ssh_fork_executor`. We can add this class 💭

pdrakeweb (approved):

lib/plugins/rotate_asg_instances/asg.rb
```diff
@@ -0,0 +1,237 @@
...
+
+    def outdated_instances
+      asg.instances.reject do |i|
+        i.launch_configuration_name == asg.launch_configuration_name
+      end
+    end
+
+    private
+
+    def outdated_volumes(instances = outdated_instances)
```
kaushik, line 49: Re raising the comment that we can skip the argument here and directly using outdated_instances inside the method.

lib/plugins/rotate_asg_instances/asg.rb
```diff
@@ -0,0 +1,237 @@
...
+    #
+    # Each instance will be detached one at a time, waiting for the new instance
+    # to be ready before stopping the worker and terminating the instance.
+    #
+    # @param instances [Array] (outdated instances)
+    #   List of instances to cycle. Defaults to all instances with outdated
+    #   launch configurations.
+    # @return [Array] (array of Aws::AutoScaling::Instance)
+    #   List of shutdown instances.
+    def cycle_instances(instances = outdated_instances)
```
kaushik, line 75: Same comment as above

lib/plugins/rotate_asg_instances/asg.rb
kaushik: It means updating the outdated_instances method too.
```ruby
        def outdated_instances
          @outdated_instances ||= asg.instances.reject do |i|
            i.launch_configuration_name == asg.launch_configuration_name
         end
       end
```

kaushik (dismissed):

lib/plugins/rotate_asg_instances_lib/system.rb
kaushik: @AbhishekRana95 It doesn't look like we are using any of the additional functionality that is present in the class you added. We should just use the method already in moonshot if that will do what we need.

AbhishekRana95: @kaushik used `ssh_fork_executor` for exec functionality.

lib/plugins/rotate_asg_instances/asg.rb
AbhishekRana95: Made the changes suggested here.

glofishh:
Manually reviewed! Verified instance was properly stopped, removed, and rotated.

**Console output**
```
|gloria.lee| |cloudservicesdev| (master)$ [2020/10/12-16:22:03]\ in ~/Desktop/Database/cloud-database-worker
  → bundle exec moonshot update -n dev-gloria1
[ ✓ ] [ 0m 0s ] Using existing KMS Key for ParameterKMS!
[ ✓ ] [ 0m 1s ] Using previous encrypted value for NewRelicLicenseKey.
[ ✓ ] [ 0m 12s ] ChangeSet moonshot-cdb-worker-dev-gloria1-1602534307 ready!
* Modify LaunchConfig (AWS::AutoScaling::LaunchConfiguration)
 - Will be replaced
 - Caused by template change (Properties: ImageId)
* Modify WorkerASG (AWS::AutoScaling::AutoScalingGroup)
 - May be replaced (Conditional)
 - Caused by LaunchConfig (ResourceReference)
Apply changes? yes
[ ✓ ] [ 0m 0s ] Executed ChangeSet moonshot-cdb-worker-dev-gloria1-1602534307 for CloudFormation Stack cdb-worker-dev-gloria1.
[ ✓ ] [ 0m 22s ] CloudFormation Stack cdb-worker-dev-gloria1 successfully updated.
[ ✓ ] [ 0m 1s ] CodeDeploy Application cdb-worker-dev-gloria1 already exists.
[ ✓ ] [ 0m 0s ] CodeDeploy CodeDeploy Deployment Group cdb-worker-dev-gloria1 already exists.
[ ✓ ] [ 0m 0s ] AutoScaling Group(s) up to capacity!
[ ✓ ] [ 0m 1s ] Uploading 'cdb-worker_1602534353_gloria.lee.tar.gz' to 'cdb-stack-resource-backups-dev' succeeded.
[ ✓ ] [ 4m 25s ] ASG instances rotated successfully!
[ ✓ ] [ 0m 2s ] Outdated instances removed successfully!
```

**In AWS console**

_Before update_
asg_new

_After update_
Outdated instance successfully shut down:
asg_deleted
New instance created/rotation completed:
asg_new

glofishh (approved):

kaushik (approved):

glofishh:
Ran `rls-test-dev` -- _failed_.
https://cloud-data.jenkins.r53.ahdev.co/job/rls-test-dev/490/

Looked into the error and it's the same one I was getting locally during manual testing when my `MOONSHOT_SSH_USER` was set to my `CDB_CI_USER`.

I think what _might_ be the problem here is that despite defining the `MOONSHOT_SSH_USER` in the moonshot repo, the `@ssh_user` is being set to the `CDB_CI_USER` during release in a couple of places in API:

- [tools>release>asg.rb](https://github.com/acquia/cloud-database-api/blob/827bbabae6fac6c84b3968f6a6d1814b0afdd532/lib/acquia/cloud/data/tools/release/asg.rb#L28).
-  [jenkins groovy for release-test](https://github.com/acquia/cloud-database-api/blob/827bbabae6fac6c84b3968f6a6d1814b0afdd532/jenkins/release-test.groovy#L46)

If this is the cause, we might need to do some additional work around getting the env variable `MOONSHOT_SSH_USER` properly exported for a release?

PR #270, opened 2020-10-22
CPD-183: Added pre-check to plugin

AbhishekRana95:
Change: minor
Purpose: feature

**Release test**: [#511](https://cloud-data.jenkins.r53.ahdev.co/job/rls-test-dev/511) ✅
**System test**: [#1081](https://cloud-data.jenkins.r53.ahdev.co/job/sys-test-dev/1081) ✅ 
**Manual Review**: [Review document](https://confluence.acquia.com/x/vwnPBg)

### Related PRs:
- [cloud-database-api](https://github.com/acquia/cloud-database-api/pull/1739)
- [cloud-database-worker](https://github.com/acquia/cloud-database-worker/pull/1020)

Added a pre-check for verifying the plugin can work before attempting to rotate AutoScalingGroup instances.

### DIFF
--- a/docs/plugins/rotate_asg_instances.md
+++ b/docs/plugins/rotate_asg_instances.md
@@ -9,12 +9,9 @@ After all outdated instances are shutdown successfully, it terminates them and r
 It allows gracefully shutting down each instance instead of terminating them and killing all the running processes.
 
 ## Configuration
-The plugin uses the ssh username specified in the `MOONSHOT_SSH_USER` or the `LOGNAME` environment variable for logging into the ASG instances to shutdown. The value should be the username with which you have the access to the instances. For example:
-```ruby
-export MOONSHOT_SSH_USER=abhishek.rana
-```
+The plugin uses config.ssh_config.ssh_user value for logging into the ASG instances to shutdown. The value should be the username with which you have the access to the instances.
 
-The plugin needs no additional configuration parameters:
+The plugin accepts additional configuration from config.ssh_config.ssh_options for ssh.
 
 ## Example
 ```ruby
--- a/lib/moonshot/ssh_config.rb
+++ b/lib/moonshot/ssh_config.rb
@@ -2,10 +2,12 @@ module Moonshot
   class SSHConfig
     attr_accessor :ssh_identity_file
     attr_accessor :ssh_user
+    attr_accessor :ssh_options
 
     def initialize
       @ssh_identity_file = ENV['MOONSHOT_SSH_KEY_FILE']
       @ssh_user = ENV['MOONSHOT_SSH_USER']
+      @ssh_options = ENV['MOONSHOT_SSH_OPTIONS']
     end
   end
 end
--- a/lib/moonshot/ssh_fork_executor.rb
+++ b/lib/moonshot/ssh_fork_executor.rb
@@ -3,18 +3,18 @@ require 'open3'
 module Moonshot
   # Run an SSH command via fork/exec.
   class SSHForkExecutor
-    Result = Struct.new(:output, :exitstatus)
+    Result = Struct.new(:output, :error, :exitstatus)
 
     def run(cmd)
-      output = StringIO.new
-
+      output = error = StringIO.new
       exit_status = nil
-      Open3.popen3(cmd) do |_, stdout, _, wt|
+      Open3.popen3(cmd) do |_, stdout, stderr, wt|
+        error << stderr.read until stderr.eof?
         output << stdout.read until stdout.eof?
         exit_status = wt.value.exitstatus
       end
 
-      Result.new(output.string.chomp, exit_status)
+      Result.new(output.string.chomp, error.string.chomp, exit_status)
     end
   end
 end
--- a/lib/plugins/rotate_asg_instances.rb
+++ b/lib/plugins/rotate_asg_instances.rb
@@ -1,16 +1,39 @@
 # frozen_string_literal: true
 
 require 'aws-sdk'
-require_relative 'rotate_asg_instances/asg'
 
 module Moonshot
   module Plugins
     # Rotate ASG instances after update.
     class RotateAsgInstances
+      include DoctorHelper
+
+      def doctor(resources)
+        @resources = resources
+        run_all_checks
+      end
+
+      def pre_update(resources)
+        @resources = resources
+        asg.verify_ssh
+      end
+
       def post_update(resources)
-        asg = ASG.new(resources)
-        asg.rotate_asg_instances
-        asg.teardown_outdated_instances
+        @resources = resources
+        asg.perform_rotation
+      end
+
+      private
+
+      def asg
+        Moonshot::RotateAsgInstances::ASG.new(@resources)
+      end
+
+      def doctor_check_ssh
+        asg.verify_ssh
+        success('Successfully opened SSH connection to an instance.')
+      rescue Moonshot::RotateAsgInstances::SSHValidationError
+        critical('SSH connection test failed, check your SSH settings')
       end
     end
   end
--- a/lib/plugins/rotate_asg_instances/asg.rb
+++ b/lib/plugins/rotate_asg_instances/asg.rb
@@ -1,238 +1,246 @@
-require 'moonshot/ssh_fork_executor'
-
 module Moonshot
-  class ASG # rubocop:disable Metrics/ClassLength
-    include Moonshot::CredsHelper
+  module RotateAsgInstances
+    class ASG # rubocop:disable Metrics/ClassLength
+      include Moonshot::CredsHelper
 
-    def initialize(resources)
-      @resources = resources
-      @ilog = @resources.ilog
-      @ssh_user = ENV['MOONSHOT_SSH_USER'] || ENV['LOGNAME']
-    end
-
-    def asg
-      @asg ||=
-        Aws::AutoScaling::AutoScalingGroup.new(name: physical_resource_id)
-    end
-
-    def rotate_asg_instances
-      @ilog.start_threaded('Rotating ASG instances...') do |step|
-        @step = step
-        @volumes_to_delete = outdated_volumes(outdated_instances)
-        @shutdown_instances = cycle_instances(outdated_instances)
-        @step.success('ASG instances rotated successfully!')
+      def initialize(resources)
+        @resources = resources
+        @ssh = Moonshot::RotateAsgInstances::SSH.new
+        @ilog = @resources.ilog
       end
-    end
 
-    def teardown_outdated_instances
-      @ilog.start_threaded('Tearing down outdated instances...') do |step|
-        @step = step
-        terminate_instances(@shutdown_instances)
-        reap_volumes(@volumes_to_delete)
-        @step.success('Outdated instances removed successfully!')
+      def perform_rotation
+        rotate_asg_instances
+        teardown_outdated_instances
       end
-    end
 
-    def physical_resource_id
-      @resources.controller.stack
-                .resources_of_type('AWS::AutoScaling::AutoScalingGroup')
-                .first.physical_resource_id
-    end
+      def verify_ssh
+        @ssh.test_ssh_connection(first_instance_id)
+      end
 
-    def outdated_instances
-      @outdated_instances ||=
+      private
+
+      def asg
+        @asg ||=
+          Aws::AutoScaling::AutoScalingGroup.new(name: physical_resource_id)
+      end
+
+      def first_instance_id
+        SSHTargetSelector.new(
+          @resources.controller.stack,
+          asg_name: Moonshot.config.ssh_auto_scaling_group_name
+        ).choose!
+      end
+
+      def rotate_asg_instances
+        @ilog.start_threaded('Rotating ASG instances...') do |step|
+          @step = step
+          outdated = identify_outdated_instances
+          @volumes_to_delete = outdated_volumes(outdated)
+          @shutdown_instances = cycle_instances(outdated)
+          @step.success('ASG instances rotated successfully!')
+        end
+      end
+
+      def teardown_outdated_instances
+        @ilog.start_threaded('Tearing down outdated instances...') do |step|
+          @step = step
+          terminate_instances(@shutdown_instances)
+          reap_volumes(@volumes_to_delete)
+          @step.success('Outdated instances removed successfully!')
+        end
+      end
+
+      def identify_outdated_instances
         asg.instances.reject do |i|
           i.launch_configuration_name == asg.launch_configuration_name
         end
-    end
+      end
 
-    private
+      def physical_resource_id
+        @resources.controller.stack
+                  .resources_of_type('AWS::AutoScaling::AutoScalingGroup')
+                  .first.physical_resource_id
+      end
 
-    def outdated_volumes(outdated_instances)
-      volumes = []
-      outdated_instances.each do |i|
-        begin
-          inst = Aws::EC2::Instance.new(id: i.id)
-          volumes << inst.block_device_mappings.first.ebs.volume_id
-        rescue StandardError => e
-          # We're catching all errors here, because failing to reap a volume
-          # is not a critical error, will not cause issues with the update.
-          @step.failure('Failed to get volumes for instance '\
-                    "#{i.instance_id}: #{e.message}")
+      def outdated_volumes(outdated_instances)
+        volumes = []
+        outdated_instances.each do |i|
+          begin
+            inst = Aws::EC2::Instance.new(id: i.id)
+            volumes << inst.block_device_mappings.first.ebs.volume_id
+          rescue StandardError => e
+            # We're catching all errors here, because failing to reap a volume
+            # is not a critical error, will not cause issues with the update.
+            @step.failure('Failed to get volumes for instance '\
+                      "#{i.instance_id}: #{e.message}")
+          end
+        end
+        volumes
+      end
+
+      # Cycle the instances in the ASG.
+      #
+      # Each instance will be detached one at a time, waiting for the new instance
+      # to be ready before stopping the worker and terminating the instance.
+      #
+      # @param instances [Array] (outdated instances)
+      #   List of instances to cycle. Defaults to all instances with outdated
+      #   launch configurations.
+      # @return [Array] (array of Aws::AutoScaling::Instance)
+      #   List of shutdown instances.
+      def cycle_instances(outdated_instances)
+        shutdown_instances = []
+
+        if outdated_instances.empty?
+          @step.success('No instances cycled!')
+          return []
+        end
+
+        @step.success("Cycling #{outdated_instances.size} " \
+                      "of #{asg.instances.size} instances in " \
+                      "#{physical_resource_id}...")
+
+        # Iterate over the instances in the stack, detaching and terminating each
+        # one.
+        outdated_instances.each do |i|
+          next if %w(Terminating Terminated).include?(i.lifecycle_state)
+
+          wait_for_instance(i)
+          detach_instance(i)
+
+          @step.success("Shutting down #{i.instance_id}")
+          shutdown_instance(i.instance_id)
+          shutdown_instances << i
+        end
+
+        @step.success('All instances cycled.')
+
+        shutdown_instances
+      end
+
+      # Waits for an instance to reach a ready state.
+      #
+      # @param instance [Aws::AutoScaling::Instance] Auto scaling instance to wait
+      #   for.
+      def wait_for_instance(instance, state = 'InService')
+        instance.wait_until(max_attempts: 60, delay: 10) do |i|
+          i.lifecycle_state == state
         end
       end
-      volumes
-    end
 
-    # Cycle the instances in the ASG.
-    #
-    # Each instance will be detached one at a time, waiting for the new instance
-    # to be ready before stopping the worker and terminating the instance.
-    #
-    # @param instances [Array] (outdated instances)
-    #   List of instances to cycle. Defaults to all instances with outdated
-    #   launch configurations.
-    # @return [Array] (array of Aws::AutoScaling::Instance)
-    #   List of shutdown instances.
-    def cycle_instances(outdated_instances)
-      shutdown_instances = []
+      # Detach an instance from its ASG. Re-attach if failed.
+      #
+      # @param instance [Aws::AutoScaling::Instance] Instance to detach.
+      def detach_instance(instance)
+        @step.success("Detaching instance: #{instance.instance_id}")
 
-      if outdated_instances.empty?
-        @step.success('No instances cycled!')
-        return []
+        # If the ASG can't be brought up to capacity, re-attach the instance.
+        begin
+          instance.detach(should_decrement_desired_capacity: false)
+          @step.success('- Waiting for the AutoScaling '\
+                       'Group to be up to capacity')
+          wait_for_capacity
+        rescue StandardError => e
+          @step.failure("Error bringing the ASG up to capacity: #{e.message}")
+          @step.failure("Attaching instance: #{instance.instance_id}")
+          reattach_instance(instance)
+          raise e
+        end
       end
 
-      @step.success("Cycling #{outdated_instances.size} " \
-                    "of #{asg.instances.size} instances in " \
-                    "#{physical_resource_id}...")
-
-      # Iterate over the instances in the stack, detaching and terminating each
-      # one.
-      outdated_instances.each do |i|
-        next if %w(Terminating Terminated).include?(i.lifecycle_state)
-
-        wait_for_instance(i)
-        detach_instance(i)
-
-        @step.success("Shutting down #{i.instance_id}")
-        shutdown_instance(i.instance_id)
-        shutdown_instances << i
-      end
-
-      @step.success('All instances cycled.')
-
-      shutdown_instances
-    end
-
-    # Waits for an instance to reach a ready state.
-    #
-    # @param instance [Aws::AutoScaling::Instance] Auto scaling instance to wait
-    #   for.
-    def wait_for_instance(instance, state = 'InService')
-      instance.wait_until(max_attempts: 60, delay: 10) do |i|
-        i.lifecycle_state == state
-      end
-    end
-
-    # Detach an instance from its ASG. Re-attach if failed.
-    #
-    # @param instance [Aws::AutoScaling::Instance] Instance to detach.
-    def detach_instance(instance)
-      @step.success("Detaching instance: #{instance.instance_id}")
-
-      # If the ASG can't be brought up to capacity, re-attach the instance.
-      begin
-        instance.detach(should_decrement_desired_capacity: false)
-        @step.success('- Waiting for the AutoScaling '\
-                     'Group to be up to capacity')
-        wait_for_capacity
-      rescue StandardError => e
-        @step.failure("Error bringing the ASG up to capacity: #{e.message}")
-        @step.failure("Attaching instance: #{instance.instance_id}")
-        reattach_instance(instance)
-        raise e
-      end
-    end
-
-    # Re-attach an instance to its ASG.
-    #
-    # @param instance [Aws::AutoScaling::Instance] Instance to re-attach.
-    def reattach_instance(instance)
-      instance.load
-      return unless instance.data.nil? \
-        || %w(Detached Detaching).include?(instance.lifecycle_state)
-
-      until instance.data.nil? || instance.lifecycle_state == 'Detached'
-        sleep 10
+      # Re-attach an instance to its ASG.
+      #
+      # @param instance [Aws::AutoScaling::Instance] Instance to re-attach.
+      def reattach_instance(instance)
         instance.load
-      end
-      instance.attach
-    end
+        return unless instance.data.nil? \
+          || %w(Detached Detaching).include?(instance.lifecycle_state)
 
-    # Terminate instances.
-    #
-    # @param instances [Array] (instances for termination)
-    #   List of instances to terminate. Defaults to all instances with outdated
-    #   launch configurations.
-    def terminate_instances(outdated_instances)
-      if outdated_instances.any?
-        @step.continue(
-          "Terminating #{outdated_instances.size} outdated instances..."
-        )
-      end
-      outdated_instances.each do |asg_instance|
-        instance = Aws::EC2::Instance.new(asg_instance.instance_id)
-        begin
+        until instance.data.nil? || instance.lifecycle_state == 'Detached'
+          sleep 10
           instance.load
-        rescue Aws::EC2::Errors::InvalidInstanceIDNotFound
-          next
+        end
+        instance.attach
+      end
+
+      # Terminate instances.
+      #
+      # @param instances [Array] (instances for termination)
+      #   List of instances to terminate. Defaults to all instances with outdated
+      #   launch configurations.
+      def terminate_instances(shutdown_instances)
+        if shutdown_instances.any?
+          @step.continue(
+            "Terminating #{shutdown_instances.size} outdated instances..."
+          )
+        end
+        shutdown_instances.each do |asg_instance|
+          instance = Aws::EC2::Instance.new(asg_instance.instance_id)
+          begin
+            instance.load
+          rescue Aws::EC2::Errors::InvalidInstanceIDNotFound
+            next
+          end
+
+          next unless %w(stopping stopped).include?(instance.state.name)
+
+          instance.wait_until_stopped
+
+          @step.continue("Terminating #{instance.instance_id}")
+          instance.terminate
+        end
+      end
+
+      def reap_volumes(volumes)
+        volumes.each do |volume_id|
+          begin
+            @step.continue("Deleting volume: #{volume_id}")
+            ec2_client(region: ENV['AWS_REGION'])
+              .delete_volume(volume_id: volume_id)
+          rescue StandardError => e
+            # We're catching all errors here, because failing to reap a volume
+            # is not a critical error, will not cause issues with the release.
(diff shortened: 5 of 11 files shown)


lib/plugins/rotate_asg_instances/asg.rb
```diff
@@ -15,6 +13,27 @@ def asg
         Aws::AutoScaling::AutoScalingGroup.new(name: physical_resource_id)
     end
 
+    def pre_check_ssh_ability
+      return false if first_ec2_instance.nil?
+
+      command_response = run_ssh_command(first_ec2_instance, "echo 'Trying to SSH'")
+      able_to_ssh = command_response.exitstatus.zero?
+      unless able_to_ssh
+        @ilog.error('Warning: Rotation Plugin aborted and will not work ' \
```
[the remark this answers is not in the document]
AbhishekRana95: @marciotoshio even if we verify on the variable level we will still need to ssh and check if the value is correct. What are your thoughts on this?

lib/plugins/rotate_asg_instances/asg.rb
```diff
@@ -223,15 +252,10 @@ def wait_for_capacity
...
-        'StrictHostKeyChecking=no'
-      ]
-      remote = "#{@ssh_user}@#{instance.public_dns_name}"
-      cmd = "'sudo shutdown -h now'"
-      remote_cmd = "ssh -o #{options.join(' -o ')} #{remote} #{cmd}"
-      SSHForkExecutor.new.run(remote_cmd)
-
+      run_ssh_command(
+        instance, 'sudo shutdown -h now',
+        ['UserKnownHostsFile=/dev/null', 'StrictHostKeyChecking=no']
```
kaushik, line 257: I think we should use the same options for all the commands here. So why dont you move these into the run ssh command method itself.

lib/plugins/rotate_asg_instances/asg.rb
```diff
@@ -1,20 +1,39 @@
...
 
     def asg
       @asg ||=
         Aws::AutoScaling::AutoScalingGroup.new(name: physical_resource_id)
     end
 
+    def try_to_ssh(raise_exception: false)
+      Retriable.retriable(base_interval: 5, tries: 3) do
+        @command_response =
+          run_ssh_command(first_ec2_instance, "'/bin/true'")
```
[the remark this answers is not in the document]
kaushik: I think an additional option ConnectTimeout=10 like https://github.com/kaushik/cloud-database-worker/blob/97ff159f151c78258356e590226adb3330421a2d/lib/moonshot/plugins/ssh.rb#L47 will be useful when just checking an ssh connection as the command wont be hanging for very long.

kaushik (changes requested):

lib/plugins/rotate_asg_instances/asg.rb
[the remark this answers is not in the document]
AbhishekRana95: Added 💡

lib/plugins/rotate_asg_instances/asg.rb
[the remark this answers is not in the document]
AbhishekRana95: @marciotoshio added the error which caused the ssh to fail and included the `exitstatus` as well to the message.

lib/plugins/rotate_asg_instances.rb
```diff
@@ -7,10 +7,40 @@ module Moonshot
...
+      def doctor_check_ssh
+        if @asg.try_to_ssh
+          success('Successfully opened SSH connection to a worker instance.')
+        else
+          critical('SSH connection test failed, check your SSH settings' \
+            ' and make sure you have the access.')
+        end
+      end
+
+      def initialize_asg(resources)
```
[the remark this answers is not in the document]
AbhishekRana95: Renamed it to `asg`

lib/plugins/rotate_asg_instances/asg.rb
```diff
@@ -1,20 +1,39 @@
...
+      @ssh_user = ssh_user
     end
 
     def asg
       @asg ||=
         Aws::AutoScaling::AutoScalingGroup.new(name: physical_resource_id)
     end
 
+    def try_to_ssh(raise_exception: false)
+      Retriable.retriable(base_interval: 5, tries: 3) do
```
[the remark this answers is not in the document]
AbhishekRana95: Moved the `Retriable` block to another function and used it directly to get the value.

lib/plugins/rotate_asg_instances/asg.rb
AbhishekRana95: Moved the options inside the `run_ssh_command` function.

lib/plugins/rotate_asg_instances/asg.rb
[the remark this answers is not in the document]
AbhishekRana95: I think `Retriable` depends on the last line executed and whether it returns true or false based on that it executes again as per the earlier implementation in [pre_deploy_validation](https://github.com/acquia/cloud-database-api/blob/master/lib/acquia/cloud/data/tools/cli/release.rb#L573-L579). Correct me if I am wrong here.

lib/plugins/rotate_asg_instances/asg.rb
```diff
@@ -1,20 +1,35 @@
...
   class ASG # rubocop:disable Metrics/ClassLength
     include Moonshot::CredsHelper
 
-    def initialize(resources)
+    def initialize(resources, ssh_user)
       @resources = resources
       @ilog = @resources.ilog
-      @ssh_user = ENV['MOONSHOT_SSH_USER'] || ENV['LOGNAME']
+      @ssh_user = ssh_user
+      puts "ssh_user: #{@ssh_user}"
```
kaushik, line 9: I wonder a puts here would affect the interactive logging format of the tool.

lib/plugins/rotate_asg_instances/asg.rb
```diff
@@ -1,20 +1,35 @@
...
     end
 
     def asg
       @asg ||=
         Aws::AutoScaling::AutoScalingGroup.new(name: physical_resource_id)
     end
 
+    def try_to_ssh(raise_exception: false)
```
kaushik, line 17: This method returns true if successful but does not return false when ssh fails(by default)

lib/plugins/rotate_asg_instances/asg.rb
```diff
@@ -48,6 +63,22 @@ def outdated_instances
...
+        @command_response = run_ssh_command(first_ec2_instance, '/bin/true')
+        @command_response.exitstatus.zero?
+      end
+    end
+
+    def run_ssh_command(instance, command)
+      remote = "#{@ssh_user}@#{instance.public_dns_name}"
+      options = ['UserKnownHostsFile=/dev/null', 'StrictHostKeyChecking=no']
+      remote_cmd = "ssh -o #{options.join(' -o ')} #{remote} " \
+                   "-o ConnectTimeout=10 #{command}"
```
kaushik, line 77: We don't want to always add ConnectTimeout option, only for doctor and pre update checks, where we test the connection. When we run shutdown, let's skip it.

lib/plugins/rotate_asg_instances/asg.rb
```diff
@@ -48,6 +63,22 @@ def outdated_instances
 
     private
 
+    def ssh_executed_successfully?
+      Retriable.retriable(base_interval: 5, tries: 3) do
+        @command_response = run_ssh_command(first_ec2_instance, '/bin/true')
+        @command_response.exitstatus.zero?
```
kaushik, line 69: Retriable gem afaik on retries when it's block raises an error. Since this is block will only return false when command return non zero exitstatus it will not be retried.

lib/plugins/rotate_asg_instances/asg.rb
AbhishekRana95: That I put for debug purpose actually to check whether CDB_CI_USER injects correctly to plugin. It is setting correctly still rls-test breaks saying failed to ssh. Looking into it.

AbhishekRana95: I will remove it 🙂

lib/plugins/rotate_asg_instances/asg.rb
kaushik: You could do.
```ruby
ssh_executed_successfully?.tap |success|
   raise "XXXXX" unless success
end
```

kaushik (changes requested):

lib/plugins/rotate_asg_instances/asg.rb
AbhishekRana95: I implemented differently. Please let me know if the approach is acceptable.

lib/plugins/rotate_asg_instances.rb
```diff
@@ -7,11 +7,41 @@ module Moonshot
...
+      def post_update(_resources)
         asg.rotate_asg_instances
         asg.teardown_outdated_instances
       end
+
+      private
+
+      def doctor_check_ssh
+        if asg.try_to_ssh
+          success('Successfully opened SSH connection to a worker instance.')
```
marciotoshio, line 35: I think "worker instance" is related to Cloud Data service, we should say only "instance"

lib/plugins/rotate_asg_instances.rb
```diff
@@ -7,11 +7,41 @@ module Moonshot
...
+        else
+          critical('SSH connection test failed, check your SSH settings' \
```
marciotoshio, line 37: Here we are hiding the actual error, should the asg.try_to_ssh raise an exception and we rescue here and add the exception message to the critical log?

lib/plugins/rotate_asg_instances.rb
```diff
@@ -7,11 +7,41 @@ module Moonshot
...
+        doctor_check_ssh
+      end
+
+      def pre_update(resources)
+        @resources = resources
+        asg.try_to_ssh(raise_exception: true)
+      end
+
+      def post_update(_resources)
         asg.rotate_asg_instances
```
marciotoshio, line 27: Looks like the asg method needs the resources, can we pass it explicit as a parameter instead setting a instance variable, I think this way we avoid bugs like that and if we still forget the error will be clear saying we need to send resources to the method.

lib/plugins/rotate_asg_instances/asg.rb
```diff
@@ -1,20 +1,36 @@
...
-
 module Moonshot
   class ASG # rubocop:disable Metrics/ClassLength
     include Moonshot::CredsHelper
 
-    def initialize(resources)
+    def initialize(resources, ssh_user)
       @resources = resources
       @ilog = @resources.ilog
-      @ssh_user = ENV['MOONSHOT_SSH_USER'] || ENV['LOGNAME']
```
marciotoshio, line 10: Here we are removing the ability to use `ENV['LOGNAME']` but not adding it anywhere else, is it ok to lose this feature?

lib/plugins/rotate_asg_instances/asg.rb
```diff
@@ -48,6 +64,35 @@ def outdated_instances
...
+
+    def build_command(instance, command, timeout_connection, ssh_debug)
+      remote_cmd = "#{@ssh_user}@#{instance.public_dns_name}"
+      command_options = 'ssh '
+      options = ['UserKnownHostsFile=/dev/null', 'StrictHostKeyChecking=no']
+      command_options += '-v ' if ssh_debug
+      command_options += '-o ConnectTimeout=10' if timeout_connection
+      "#{command_options} -o #{options.join(' -o ')} #{remote_cmd} #{command}"
+    end
+
```
marciotoshio, line 95: All this ssh code here is to run a specific command `/bin/true` which seems a test command which makes me feel that this code does not belong to this ASG class, but to other specific to execute tests/checks

lib/plugins/rotate_asg_instances/asg.rb
```diff
@@ -48,6 +64,35 @@ def outdated_instances
...
+    def run_ssh_command(instance, command, timeout_connection: true,
+                        ssh_debug: false)
+
+      remote_cmd =
+        build_command(instance, command, timeout_connection, ssh_debug)
+      Moonshot::SSHForkExecutor.new.run(remote_cmd)
+    end
+
+    def build_command(instance, command, timeout_connection, ssh_debug)
```
marciotoshio, line 87: There's a SSHCommandBuilder class `lib/moonshot/ssh_command_builder.rb` which does things that some methods of this class is doing

marciotoshio (changes requested):

lib/plugins/rotate_asg_instances/asg.rb
kaushik: I think we should specify LOGNAME along with CDB_CI_USER in the worker PR where this plugin is configured.

lib/plugins/rotate_asg_instances/asg.rb
AbhishekRana95: @marciotoshio the `command_builder` does not support `options` or `timeout_connection` to pass and whether to use verbose which the pre-check needs. It is better if we maintain our original implementation here.

lib/plugins/rotate_asg_instances/asg.rb
AbhishekRana95: Moving it to a new module `ASGChecks`

lib/plugins/rotate_asg_instances.rb
AbhishekRana95: In case of doctor check, we do not raise an exception in the `execute_ssh` and instead we receive `true/false` from the method and raise a custom exception in this block.

lib/plugins/rotate_asg_instances.rb
AbhishekRana95: Made it a parameter and removed the instance variable implementation.

lib/moonshot/ssh_fork_executor.rb
```diff
@@ -9,7 +9,8 @@ def run(cmd)
       output = StringIO.new
 
       exit_status = nil
-      Open3.popen3(cmd) do |_, stdout, _, wt|
+      Open3.popen3(cmd) do |_, stdout, stderr, wt|
```
kaushik, line 12: Adding this makes this PR modify moonshot internals and not just add a new plugin. Not sure if we want that.
Main concern being it makes makes merging into master when we can little more complicated compared to only adding a plugin.

pdrakeweb: I'm not sure we just want to add stderr to the "output" property of the result object.  It would probably be better to modify the Result struct to add stderr as a separate property.

pdrakeweb (changes requested):

lib/plugins/rotate_asg_instances/asg_checks.rb
```diff
@@ -0,0 +1,67 @@
...
+        @command_response.exitstatus.zero?
+      end
+    end
+
+    def first_ec2_instance
+      first_instance = asg.instances.first
+      return if first_instance.nil?
+
+      Aws::EC2::Instance.new(id: first_instance.instance_id)
+    end
```
marciotoshio, line 65: I think we should use the classes that moonshot have to ssh into the instance, here we are just adding a few options that I think we don't want to add, like ignoring the strict host key check. Another point is that the builder command class already has code to select an instance and we don't need to duplicate this in this class.
And then maybe we can just call the Moonshot ssh classes from ASG rotate plugin class.

kaushik: Makes sense

kaushik:
@AbhishekRana95 can you please resolve outdated comments if they have been addressed.

lib/moonshot/ssh_fork_executor.rb
AbhishekRana95: Thank you for the suggestion @pdrakeweb

lib/plugins/rotate_asg_instances/asg_checks.rb
```diff
@@ -0,0 +1,67 @@
...
+    def build_command(instance, command, timeout_connection)
+      remote_cmd = "#{@ssh_user}@#{instance.public_dns_name}"
+      command_options = 'ssh '
+      options = ['UserKnownHostsFile=/dev/null', 'StrictHostKeyChecking=no']
+      command_options += '-v ' if print_ssh_debug?
+      command_options += '-o ConnectTimeout=10' if timeout_connection
+      "#{command_options} -o #{options.join(' -o ')} #{remote_cmd} #{command}"
+    end
+
+    def execute_ssh
```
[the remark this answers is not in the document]
AbhishekRana95: Makes sense.

spec/moonshot/plugins/asg.rb
```diff
@@ -282,4 +267,48 @@ def stub_cf_client
       end
     end
   end
+
+  describe '#pre_check_ssh_ability' do
```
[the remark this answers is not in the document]
AbhishekRana95: Yes we can

docs/plugins/rotate_asg_instances.md
```diff
@@ -9,17 +9,22 @@ After all outdated instances are shutdown successfully, it terminates them and r
 It allows gracefully shutting down each instance instead of terminating them and killing all the running processes.
 
 ## Configuration
-The plugin uses the ssh username specified in the `MOONSHOT_SSH_USER` or the `LOGNAME` environment variable for logging into the ASG instances to shutdown. The value should be the username with which you have the access to the instances. For example:
-```ruby
-export MOONSHOT_SSH_USER=abhishek.rana
-```
+The plugin uses the ssh configuration for logging into the ASG instances to shutdown. The value should be the username with which you have the access to the instances.
```
pdrakeweb, line 12: `The value` here is unclear.  What value?   I presume ssh_config.ssh_user based on the block below but this sentence is unclear.

docs/plugins/rotate_asg_instances.md
```diff
@@ -9,17 +9,22 @@ After all outdated instances are shutdown successfully, it terminates them and r
...
 
-The plugin needs no additional configuration parameters:
+The plugin accepts additional configuration parameters for ssh which should be an array:
 
 ## Example
 ```ruby
 Moonshot.config do |c|
   # ...
+  # Environment setup for ASG rotation plugin
+  c.ssh_config.ssh_user = 'abhishek.rana'
```
pdrakeweb, line 21: I don't think this is a particularly valid example as I'm not sure that anyone is going to hard-code an SSH username.  Is an environment variable a valid option?  Can we show that?

lib/moonshot/ssh_fork_executor.rb
```diff
@@ -3,18 +3,18 @@
...
+    Result = Struct.new(:output, :error, :exitstatus)
 
     def run(cmd)
-      output = StringIO.new
-
+      output = error = StringIO.new
       exit_status = nil
-      Open3.popen3(cmd) do |_, stdout, _, wt|
+      Open3.popen3(cmd) do |_, stdout, stderr, wt|
+        error << stderr.read
```
pdrakeweb, line 12: Does this need to be `error << stderr.read until stderr.eof?` similar to stdout below?

lib/moonshot/ssh_config.rb
```diff
@@ -2,10 +2,12 @@ module Moonshot
   class SSHConfig
     attr_accessor :ssh_identity_file
     attr_accessor :ssh_user
+    attr_accessor :ssh_optional_parameters
```
kaushik, line 5: This can just be :ssh_options

lib/moonshot/ssh_config.rb
```diff
@@ -2,10 +2,12 @@ module Moonshot
...
 
     def initialize
       @ssh_identity_file = ENV['MOONSHOT_SSH_KEY_FILE']
       @ssh_user = ENV['MOONSHOT_SSH_USER']
+      @ssh_optional_parameters = []
```
kaushik, line 10: Maybe this can just be a string, and we can define a default environment variable for it MOONSHOT_SSH_OPTIONS?

kaushik: And does it have to be an array, can it just be a string, for example "-v -o StrictHostKeyChecking=no". Might help with easy configuration from environment variable or directly passing the string.

lib/plugins/rotate_asg_instances.rb
```diff
@@ -1,16 +1,50 @@
...
+        auto_scaling_group.rotate_asg_instances
+        auto_scaling_group.teardown_outdated_instances
+      end
+
+      def asg(resources)
+        @asg ||=
+          Aws::AutoScaling::AutoScalingGroup.new(name: physical_resource_id(resources))
+      end
+
+      def asg_ssh_class(resources)
```
kaushik, line 31: I don't think this method should say class in it as it does not return a class.
I think it can just be ssh, as the context that it is asg plugin related object is clear in its definition.

lib/plugins/rotate_asg_instances.rb
```diff
@@ -1,16 +1,50 @@
...
+      private
+
+      def doctor_check_ssh(resources)
+        asg_ssh_class(resources).try_to_ssh
+        success('Successfully opened SSH connection to an instance.')
+      rescue Moonshot::Asg::SSHValidationError
+        critical('SSH connection test failed, check your SSH settings')
+      end
+
+      def physical_resource_id(resources)
```
kaushik, line 44: Why was this method moved out of asg class? I think it belongs there.

lib/plugins/rotate_asg_instances/ssh.rb
```diff
@@ -0,0 +1,44 @@
...
+      def test_ssh_connection
+        Retriable.retriable(base_interval: 5, tries: 3) do
+          @command_response = exec('/bin/true')
+          raise SSHValidationError, 'Validation failed.' unless
+            @command_response.exitstatus.zero?
+          @command_response.exitstatus.zero?
+        end
+      end
+
+      def first_instance_id(resources)
```
kaushik, line 36: I think this method belongs to asg class.

lib/plugins/rotate_asg_instances.rb
```diff
@@ -1,16 +1,50 @@
...
-        asg = ASG.new(resources)
-        asg.rotate_asg_instances
-        asg.teardown_outdated_instances
+        auto_scaling_group = Moonshot::ASG.new(resources, asg(resources),
+                                               physical_resource_id(resources))
+        auto_scaling_group.rotate_asg_instances
+        auto_scaling_group.teardown_outdated_instances
+      end
+
+      def asg(resources)
```
kaushik, line 26: Why not this just be asg class like before and that will handle looking up aws for physical resource id, getting the instances.

and for testing ssh we could do try_ssh(asg.first_instance)

docs/plugins/rotate_asg_instances.md
kaushik: I don't think we need to specify any of these examples, as ssh config here is not part of the plugin. Its upto the moonshot user on how they configure their ssh using ssh_config.

lib/moonshot/ssh_config.rb
AbhishekRana95: That's what we discussed last time. It can be a string and yes, we can convert it to an ENV variable.

lib/plugins/rotate_asg_instances.rb
AbhishekRana95: Moved back

lib/plugins/rotate_asg_instances/ssh.rb
AbhishekRana95: Moved back

lib/moonshot/ssh_config.rb
AbhishekRana95: I converted it to ENV. I can see in `/spec/moonshot/ssh_spec.rb` the value is being used now, and the Moonshot core ssh is getting the new value.

AbhishekRana95: The valid solution to this is I removed initialization from [here](https://github.com/acquia/moonshot/pull/270/files#diff-02d46b8526660e3a96b20d4108ece8e462599ab08f0dac29f15845a2725a458fR7) for the `attr_accessor` and just inject the value through [cdb-worker Moonfile.rb](https://github.com/acquia/cloud-database-worker/pull/1020/files#diff-9a6c7ec0145dc4264f61b3a8be25fb874d9bd82b1fa35a53a7eb1a194ff2747bR117)

docs/plugins/rotate_asg_instances.md
AbhishekRana95: Removed the example.

lib/moonshot/ssh_fork_executor.rb
AbhishekRana95: Thanks for the tip!

lib/moonshot/ssh_config.rb
AbhishekRana95: Renamed to :ssh_options

lib/plugins/rotate_asg_instances.rb
AbhishekRana95: Renamed to ssh.

lib/plugins/rotate_asg_instances.rb
AbhishekRana95: Moved methods back to `asg` and sending `first_instance_id` from `asg` to `ssh` class.

spec/moonshot/plugins/asg_spec.rb
```diff
@@ -108,64 +118,35 @@ def stub_cf_client
...
-    it 'runs SSH with proper option to ignore host keys' do
-      ENV['MOONSHOT_SSH_USER'] = 'ci_user'
-      opts_string = '-o UserKnownHostsFile=/dev/null ' \
-                    '-o StrictHostKeyChecking=no'
-      expect(System).to receive(:exec).with(
-        /#{opts_string}/,
-        any_args
+    it 'issues a shutdown with options to the instance' do
+      moonshot_config.ssh_config.ssh_options =
+        '-v -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o ConnectTimeout=10'
```
kaushik, line 145: We don't have to provide all these options here. Just provide "-v  -o userknownhostsfile"

kaushik:
We generally mention related PRs in the description so that its easy to quickly jump to them and check.

kaushik (approved):
Looks good. make sure travis passes too. And we would need a manual review.

spec/moonshot/plugins/asg_spec.rb
```diff
@@ -2,20 +2,32 @@
   let(:name) { 'cdb-worker-dev-jarmes-WorkerASG-Q7DYM7901RBY' }
```
marciotoshio, line 2: Can we follow the class file path here on the spec `spec/moonshot/plugins/rotate_asg_instances/asg_spec.rb `

spec/moonshot/plugins/ssh_spec.rb
```diff
@@ -0,0 +1,50 @@
+describe Moonshot::Asg::SSH do
```
marciotoshio, line 1: Can we follow the class file path here on the spec `spec/moonshot/plugins/rotate_asg_instances/ssh_spec.rb `

lib/plugins/rotate_asg_instances/asg.rb
```diff
@@ -1,25 +1,42 @@
-require 'moonshot/ssh_fork_executor'
-
 module Moonshot
   class ASG # rubocop:disable Metrics/ClassLength
```
marciotoshio, line 2: Should we have a module RotateAsgInstances to follow the folder path and give a better namespace to this class?

lib/plugins/rotate_asg_instances/ssh.rb
```diff
@@ -0,0 +1,34 @@
+module Moonshot
+  module Asg
```
marciotoshio, line 2: Should the module be RotateAsgInstances?

marciotoshio (changes requested):

spec/moonshot/ssh_spec.rb
```diff
@@ -6,7 +6,7 @@
     c.ssh_config.ssh_user = 'joeuser'
     c.ssh_config.ssh_identity_file = '/Users/joeuser/.ssh/thegoods.key'
     c.ssh_command = 'cat /etc/passwd'
-
+    c.ssh_config.ssh_options = nil
```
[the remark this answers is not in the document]
AbhishekRana95: Setting it to `nil` because the current test cases do not expect any options, but if you have set it on your machine, the test cases breaks.

lib/plugins/rotate_asg_instances/ssh.rb
AbhishekRana95: Makes sense

spec/moonshot/plugins/ssh_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    	allow(subject).to receive(:test_ssh_connection).with(instance_id).and_return(successful_response)
+    	expect { subject.test_ssh_connection(instance_id) }.not_to raise_error
+    end
+  end
+
+  describe '#exec' do
+  	it 'executes the command given' do
+  		allow(subject).to receive(:exec).with(command, instance_id).and_return(successful_response)
+  		expect(subject.exec(command, instance_id)).to eql(successful_response)
+  	end
```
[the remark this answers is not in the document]
AbhishekRana95: Indentation looks correct when viewing on my machine. Weird.

lib/plugins/rotate_asg_instances/asg.rb
AbhishekRana95: Makes sense.

marciotoshio:
@AbhishekRana95 Could you please check your editor tab configuration, about the wrong indentation, I think your editor is using tabs instead spaces. That's is what I see on my editor dot means space dash means tab.
Screen Shot 2020-11-06 at 13 28 00

marciotoshio (approved):

AbhishekRana95:
@marciotoshio replaced tabs with spaces.

spec/moonshot/plugins/rotate_asg_instances/asg_spec.rb
```diff
@@ -108,64 +112,34 @@ def stub_cf_client
...
+      allow(instance).to receive(:wait_until_stopped)
     end
 
-    it 'issues a shutdown to the instance' do
-      ENV['MOONSHOT_SSH_USER'] = 'ci_user'
-      expect(System).to receive(:exec).with(
-        /ssh (.*) ci_user@#{hostname} 'sudo shutdown -h now'/,
-        raise_on_failure: false
+    it 'issues a shutdown without options to the instance' do
+      moonshot_config.ssh_config.ssh_options = nil
```
pdrakeweb, line 130: Can this be moved to a `let` statement instead?

spec/moonshot/plugins/rotate_asg_instances/asg_spec.rb
```diff
@@ -108,64 +112,34 @@ def stub_cf_client
...
 
-    it 'runs SSH with proper option to ignore host keys' do
-      ENV['MOONSHOT_SSH_USER'] = 'ci_user'
-      opts_string = '-o UserKnownHostsFile=/dev/null ' \
-                    '-o StrictHostKeyChecking=no'
-      expect(System).to receive(:exec).with(
-        /#{opts_string}/,
-        any_args
+    it 'issues a shutdown with options to the instance' do
+      moonshot_config.ssh_config.ssh_options =
```
pdrakeweb, line 138: Same as above, this be moved to a `let` statement instead?

spec/moonshot/ssh_spec.rb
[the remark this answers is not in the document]
pdrakeweb: What do you think about mocking the ENV variable for the specific test instead of setting this to nil here?

pdrakeweb (approved):

spec/moonshot/plugins/rotate_asg_instances/asg_spec.rb
AbhishekRana95: Moved

spec/moonshot/plugins/rotate_asg_instances/asg_spec.rb
AbhishekRana95: Moved 🙂

spec/moonshot/plugins/rotate_asg_instances/asg_spec.rb
```diff
@@ -108,64 +112,37 @@ def stub_cf_client
...
-    end
-  end
-
   describe '#shutdown_instance' do
-    let(:hostname) { 'ec2-54-236-102-14.compute-1.amazonaws.com' }
+    let(:public_ip_address) { '10.234.32.21' }
     let(:instance) { instance_double(Aws::EC2::Instance) }
+    let(:command_builder) { Moonshot::SSHCommandBuilder }
     subject { super().send(:shutdown_instance, instance_id) }
+    let(:set_ssh_options_to_nil) { moonshot_config.ssh_config.ssh_options = nil }
```
pdrakeweb, line 120: Generally, the use of `let` should be declarative, not procedural.  In this case, it might look something like this:

```
let(:ssh_options) { nil }

before_each do
  moonshot_config.ssh_config.ssh_options = ssh_options
end

it 'issues a shutdown without options...' do
  ...
end

context 'when SSH options are defined'
  let(:ssh_optons) { '-v' }
  
  it 'issues a shutdown with options' do
    ...
  end
end
```

AbhishekRana95: I will correct it.

AbhishekRana95: I corrected. Please check.

PR #276, opened 2021-09-24
CL-6573: Unpin dependency gems.

pdrakeweb:


lib/moonshot/artifact_repository/s3_bucket.rb
```diff
@@ -1,3 +1,9 @@
+# frozen_string_literal: true
+
+require_relative '../resources_helper'
+require_relative '../creds_helper'
+require_relative '../doctor_helper'
```
kaushik, line 5: Is this a ruby 2.7 or rubocop requirement to require all libraries?

pdrakeweb: This is due to an interaction between the latest require_all gem and ruby 2.7 I believe.  Hard to remember since I wrote it a while ago.

PR #278, opened 2021-11-19
CPD-6722: Stop checking terminated instances for attached volumes.

pdrakeweb:


kaushik (approved):
Looks good to merge.

Do you think we should log the instance that was initially in the group and then not included cos its was already detached on https://github.com/acquia/moonshot/blob/8868f8bd48e78956ce2227cf1245949a95daad2e/lib/plugins/rotate_asg_instances/asg.rb#L185?

marciotoshio (approved):

pdrakeweb:
@kaushik I don't know of a case where that log info would be used since this is catching an expected scenario and continuing.  We would just be logging an expected condition that we won't look at I believe.

PR #288, opened 2022-10-25
CPD-7518: Add tags option to update command.

marciotoshio:


lib/moonshot/commands/tag_arguments.rb
```diff
@@ -0,0 +1,18 @@
+module Moonshot
+  module Commands
+    module TagArguments
+      def parser
+        parser = super
+
+        parser.on('--tag KEY=VALUE', '-TKEY=VALUE', 'Specify Stack Tag on the command line') do |v|
```
kaushik, line 7: Can we add something to describe how to specify multiple tags?

Also did you prefer using option multiple times vs using --tags once and giving it comma seperated list of key values?

marciotoshio: I don't want to change how it works currently, we have --parameter options that works exactly the same making it work like this will add complexity in the usage as we need to educate on the new usage

kaushik: Okay, we could still say ", for multiple tags repeat the tag option"

kaushik (approved):

PR #290, opened 2022-11-30
CPD-7592 keep standard tags when updating

lethanh273:
Change: patch
Purpose: maintenance

Manual testing:
in cloud-database-api folder, replace moonshot with 
```
  gem 'moonshot',      git: 'git@github.com:lethanh273/moonshot.git',      ref: 'master' 
```

bundle update moonshot 
cd network 
bundle exec moonshot update -n dev-ltt
rm cloud_formation/cdb-network.json if need to
The stack tags will be kept as they are.

### DIFF
--- a/lib/moonshot/stack.rb
+++ b/lib/moonshot/stack.rb
@@ -39,6 +39,7 @@ module Moonshot
       raise "No stack found #{@name.blue}!" unless stack_exists?
 
       change_set = ChangeSet.new(new_change_set, @name)
+
       wait_for_change_set(change_set)
       return unless change_set.valid?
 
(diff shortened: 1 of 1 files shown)


kaushik (changes requested):
None of cloud data specific information should be in moonshot. This is an independent gem used to manage cloud formation deployments.

lib/moonshot/stack.rb
```diff
@@ -301,7 +309,7 @@ def make_tags
         default_tags << { key: @config.additional_tag, value: @name }
       end
 
-      default_tags + @config.extra_tags
+      (default_tags + @config.extra_tags + existing_tags).uniq!
```
kaushik, line 312: The issue I see here is that even when we do want to remove the tags this will always keep the existing tags.

PR #291, opened 2023-02-13
Update for Ruby 3.1.3

stephenthedev:


anavarre:
Thanks for that. This is going to unblock Pipelines with the Ruby 2.7 EOL. Who can review and merge? Is this only @jfarrell ?

imnetworku (approved):
Looks good to me, but a review from someone more familiar with it would be a recommendation. Thanks a lot for the great work @stephenthedev

PR #292, opened 2023-02-24
CPD-7699: Remove Travis.

justinbm4:
Change: patch
Purpose: maintenance


kaushik (changes requested):
Looking at https://github.com/acquia/moonshot/search?q=travis
we would need to remove the documentation from build.md and travis badge from readme.md

PR #293, opened 2023-06-02
CPD-7696: adding tags for kms keys.

justinbm4:
Change: patch
Purpose: maintenance

### DIFF
--- a/lib/moonshot/stack.rb
+++ b/lib/moonshot/stack.rb
@@ -12,10 +12,30 @@ module Moonshot
     attr_reader :app_name
     attr_reader :name
 
+    class << self
+      def generate_name(config)
+        [config.app_name, config.environment_name].join('-')
+      end
+
+      def make_tags(config)
+        default_tags = [
+          { key: 'moonshot_application', value: config.app_name },
+          { key: 'moonshot_environment', value: config.environment_name },
+        ]
+        name = generate_name(config)
+
+        if config.additional_tag
+          default_tags << { key: config.additional_tag, value: name }
+        end
+
+        default_tags + config.extra_tags
+      end
+    end
+
     def initialize(config)
       @config = config
       @ilog = config.interactive_logger
-      @name = [@config.app_name, @config.environment_name].join('-')
+      @name = self.class.generate_name(@config)
 
       yield @config if block_given?
     end
@@ -292,16 +312,7 @@ module Moonshot
     end
 
     def make_tags
-      default_tags = [
-        { key: 'moonshot_application', value: @config.app_name },
-        { key: 'moonshot_environment', value: @config.environment_name }
-      ]
-
-      if @config.additional_tag
-        default_tags << { key: @config.additional_tag, value: @name }
-      end
-
-      default_tags + @config.extra_tags
+      self.class.make_tags(@config)
     end
 
     def format_event(event)
--- a/lib/plugins/encrypted_parameters.rb
+++ b/lib/plugins/encrypted_parameters.rb
@@ -102,7 +102,9 @@ module Moonshot
               Moonshot.config.parameters[@kms_key_parameter_name].set(key_arn)
               s.success "Created a new KMS Key for #{@kms_key_parameter_name.blue}!"
             else
-              key_arn = KmsKey.new(Moonshot.config.parameters[@kms_key_parameter_name].value).arn
+              kms=KmsKey.new(Moonshot.config.parameters[@kms_key_parameter_name].value)
+              key_arn = kms.arn
+              kms.update
               s.success "Using existing KMS Key for #{@kms_key_parameter_name.blue}!"
             end
           end
--- a/lib/plugins/encrypted_parameters/kms_key.rb
+++ b/lib/plugins/encrypted_parameters/kms_key.rb
@@ -1,20 +1,38 @@
+require_relative '../../moonshot/stack.rb'
+
 module Moonshot
   module Plugins
     class EncryptedParameters
       # Class that manages KMS keys in AWS.
       class KmsKey
         attr_reader :arn
+        class << self
+          def create
+            standard_tags = stack_tags
+            resp = Aws::KMS::Client.new.create_key({
+               tags:  standard_tags, # An array of tags.
+              })
+            arn = resp.key_metadata.arn
+            new(arn)
+          end
+
+          def stack_tags
+            tags = Moonshot::Stack.make_tags(Moonshot.config)
+            tags.map { |tag| { tag_key:  tag[:key], tag_value:  tag[:value] } }
+          end
+        end
 
         def initialize(arn)
           @arn = arn
           @kms_client = Aws::KMS::Client.new
         end
 
-        def self.create
-          resp = Aws::KMS::Client.new.create_key
-          arn = resp.key_metadata.arn
-
-          new(arn)
+        def update
+          standard_tags = self.class.stack_tags
+          @kms_client.tag_resource({
+            key_id: @arn, # arn of the CMK being tagged
+            tags: standard_tags, # An array of tags.
+          })
         end
 
         def delete


justinbm4:
manual testing: key tagged!
image

lib/moonshot/stack.rb
```diff
@@ -291,17 +291,22 @@ def wait_for_stack_state(wait_target, past_tense_verb)
...
 
-    def make_tags
+    def self.make_tags(config)
       default_tags = [
-        { key: 'moonshot_application', value: @config.app_name },
-        { key: 'moonshot_environment', value: @config.environment_name }
+        { key: 'moonshot_application', value: config.app_name },
+        { key: 'moonshot_environment', value: config.environment_name },
       ]
+      name=[config.app_name, config.environment_name].join('-')
```
kaushik, line 299: line 18 and this are duplicated. To maintain DRY, I think this should be wrapped in its own method.
If someone changes line 18 and misses this line that creates inconsistencies.

lib/plugins/encrypted_parameters/kms_key.rb
```diff
@@ -13,10 +15,24 @@ def initialize(arn)
...
+          resp = @kms_client.tag_resource({
+            key_id: @arn, # arn of the CMK being tagged
+            tags: standard_tags, # A list of tags.
+          })
+        end
+
+        def get_tags
+          tags=Moonshot::Stack.make_tags(Moonshot.config)
+          formated_tags=[]
+          tags.each{ |tag| formated_tags << { tag_key: tag[:key], tag_value: tag[:value]}}
```
kaushik, line 32: These three lines can be rewritten as tags.map { |tag| { tag_key: tag[:key], tag_value: tag[:value]} }

lib/plugins/encrypted_parameters/kms_key.rb
```diff
@@ -13,10 +15,24 @@ def initialize(arn)
...
 
+        def update
+          standard_tags=get_tags
+          resp = @kms_client.tag_resource({
+            key_id: @arn, # arn of the CMK being tagged
+            tags: standard_tags, # A list of tags.
+          })
+        end
+
+        def get_tags
```
kaushik, line 29: Can this be other descriptive name, something like stack_tags instead? and usually get* set* methods are used as accessor methods on the class object itself in ruby.

lib/plugins/encrypted_parameters.rb
```diff
@@ -102,9 +102,11 @@ def find_or_create_kms_key
               Moonshot.config.parameters[@kms_key_parameter_name].set(key_arn)
               s.success "Created a new KMS Key for #{@kms_key_parameter_name.blue}!"
             else
-              key_arn = KmsKey.new(Moonshot.config.parameters[@kms_key_parameter_name].value).arn
+              kms=KmsKey.new(Moonshot.config.parameters[@kms_key_parameter_name].value)
+              key_arn = kms.arn
               s.success "Using existing KMS Key for #{@kms_key_parameter_name.blue}!"
             end
+            KmsKey.new(key_arn).update
```
kaushik, line 109: I think having tags included in creation makes one api call instead of two to aws. In the current approach its optimized to have single method. I suggest we make fewer api call as each call makes it take time and more exposed to api based errors.

kaushik (changes requested):

lib/moonshot/stack.rb
justinbm4: had to make it a class method instead of an instance, but did it

lib/plugins/encrypted_parameters.rb
justinbm4: had to make stack_tags a class method for it to work, but did it.

justinbm4:
manual testing of new changes
image
image

lib/moonshot/stack.rb
```diff
@@ -291,17 +295,22 @@ def wait_for_stack_state(wait_target, past_tense_verb)
...
 
-    def make_tags
+    def self.make_tags(config)
       default_tags = [
-        { key: 'moonshot_application', value: @config.app_name },
-        { key: 'moonshot_environment', value: @config.environment_name }
+        { key: 'moonshot_application', value: config.app_name },
+        { key: 'moonshot_environment', value: config.environment_name },
       ]
+      name=Stack.generate_name(config)
```
kaushik, line 303: You can skip the Stack cal as you are calling a class method from a class method. nit: space around '=''.
```suggestion
      name = generate_name(config)
```

lib/moonshot/stack.rb
```diff
@@ -291,17 +295,22 @@ def wait_for_stack_state(wait_target, past_tense_verb)
...
+      if config.additional_tag
+        default_tags << { key: config.additional_tag, value: name }
       end
 
-      default_tags + @config.extra_tags
+      default_tags + config.extra_tags
+    end
+
+    def make_tags
+      Stack.make_tags(@config)
```
kaushik, line 313: Here, conventions is to use self.class to call a method in the same class

lib/plugins/encrypted_parameters.rb
```diff
@@ -102,8 +102,10 @@ def find_or_create_kms_key
               Moonshot.config.parameters[@kms_key_parameter_name].set(key_arn)
               s.success "Created a new KMS Key for #{@kms_key_parameter_name.blue}!"
             else
-              key_arn = KmsKey.new(Moonshot.config.parameters[@kms_key_parameter_name].value).arn
+              kms=KmsKey.new(Moonshot.config.parameters[@kms_key_parameter_name].value)
+              key_arn = kms.arn
               s.success "Using existing KMS Key for #{@kms_key_parameter_name.blue}!"
+              KmsKey.new(key_arn).update
```
kaushik, line 108: Why do we need to create a new object again here? could it just be kms.update since kms contains the KmsKey object.

lib/plugins/encrypted_parameters/kms_key.rb
```diff
@@ -11,12 +13,27 @@ def initialize(arn)
         end
 
         def self.create
-          resp = Aws::KMS::Client.new.create_key
+          standard_tags=KmsKey.stack_tags
```
kaushik, line 16: same, class method can be directly called.
```suggestion
          standard_tags = stack_tags
```

lib/plugins/encrypted_parameters/kms_key.rb
```diff
@@ -11,12 +13,27 @@ def initialize(arn)
...
 
+        def update
+          standard_tags=KmsKey.stack_tags
+          resp = @kms_client.tag_resource({
+            key_id: @arn, # arn of the CMK being tagged
+            tags: standard_tags, # An array of tags.
+          })
+        end
+
+        def self.stack_tags
```
kaushik, line 32: https://railsware.com/blog/better-ruby-choosing-convention-for-class-methods-definition/, 
how about you just follow this convention.
Put al class methods in the class << self block. It will be more intuitive.

lib/plugins/encrypted_parameters/kms_key.rb
```diff
@@ -11,12 +13,27 @@ def initialize(arn)
...
+          resp = Aws::KMS::Client.new.create_key({
+            tags: standard_tags, # An array of tags.
+          })
           arn = resp.key_metadata.arn
-
           new(arn)
         end
 
+        def update
+          standard_tags=KmsKey.stack_tags
```
kaushik, line 25: can be self.class.stack_tags

https://stackoverflow.com/a/12857426 has a good example on one reason why this is a convention.

kaushik (changes requested):

justinbm4:
manual testing of most recent changes
image
image

justinbm4:
testing update tagging
api key before update:
image
api key after update:

lib/plugins/encrypted_parameters.rb
```diff
@@ -102,8 +102,10 @@ def find_or_create_kms_key
               Moonshot.config.parameters[@kms_key_parameter_name].set(key_arn)
               s.success "Created a new KMS Key for #{@kms_key_parameter_name.blue}!"
             else
-              key_arn = KmsKey.new(Moonshot.config.parameters[@kms_key_parameter_name].value).arn
+              kms=KmsKey.new(Moonshot.config.parameters[@kms_key_parameter_name].value)
+              key_arn = kms.arn
               s.success "Using existing KMS Key for #{@kms_key_parameter_name.blue}!"
+              kms.update
```
kaushik, line 108: Can we move this line above the success log before it

kaushik (approved):

stephenthedev (approved):

PR #310, opened 2024-03-29
CPD-8525 | When in US regions, the api and worker should use AWS FIPS endpoints where available

suvrat-joshi:
**Motivation**
FIPS endpoint is only supported from aws-sdk v3 onwards so we have to upgrade the gem from v2 to v3. 

**Proposed changes**
Updates aws-sdk gem to v3

**Alternatives considered**
N/A

**Testing steps**
N/A

### DIFF
--- a/lib/moonshot.rb
+++ b/lib/moonshot.rb
@@ -1,7 +1,13 @@
 # frozen_string_literal: true
 
 require 'English'
-require 'aws-sdk'
+
+require 'aws-sdk-cloudformation'
+require 'aws-sdk-codedeploy'
+require 'aws-sdk-ec2'
+require 'aws-sdk-iam'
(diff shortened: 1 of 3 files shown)


shrikantbhosaleacquia (approved):

stephenthedev (commented):
This doesn't make us use the fips endpoints?

suvrat-joshi:
> This doesn't make us use the fips endpoints?

No, we need aws-sdk v3 for ruby for the FIPS endpoint, and this changes is for having moonshot compatible with the aws-sdk-ruby v3. Since all these are part of CPD-8525 ticket, hence I added it to the PR title. But this doesn't mandate FIPS endpoint.

stephenthedev (approved):

PR #323, opened 2025-07-01
CPD-11886: Upload template to s3 before validation.

rutuja810:
Change: minor
Purpose: maintenance

**Motivation**

Fixes #NNN

**Proposed changes**

**Alternatives considered**

**Testing steps**

### DIFF
--- a/lib/moonshot/dynamic_template.rb
+++ b/lib/moonshot/dynamic_template.rb
@@ -58,10 +58,9 @@ module Moonshot
     end
 
     def validate_template(template)
-      if template.bytesize > 50_000 # Leave some margin from the 51,200 limit
+      if Moonshot.config.template_s3_bucket
         s3_client = Aws::S3::Client.new
         bucket_name = "#{Moonshot.config.template_s3_bucket}"
-        raise ArgumentError, "template_s3_bucket configuration is required for large templates" unless bucket_name
         template_key = "#{Moonshot.config.app_name}/#{Moonshot.config.environment_name}/#{File.basename(@destination)}"
 
         # Ensure bucket exists


lib/moonshot/dynamic_template.rb
```diff
@@ -58,9 +58,37 @@ def validate_destination_exists
     end
 
     def validate_template(template)
-      Aws::CloudFormation::Client.new.validate_template(
-        template_body: template
-      )
+      if template.bytesize > 50_000 # Leave some margin from the 51,200 limit
+        s3_client = Aws::S3::Client.new
+        bucket_name = "#{Moonshot.config.template_s3_bucket}"
```
kaushik, line 63: if this returns empty or nil because template_s3_bucket we should raise an error

kaushik (approved):

lib/moonshot/stack.rb
```diff
@@ -208,7 +208,7 @@ def get_stack(name)
     def upload_template_to_s3
       raise 'The S3 bucket to store the template in is not configured.' unless @config.template_s3_bucket
 
-      s3_object_key = "#{@name}-#{Time.now.getutc.to_i}-#{File.basename(template.filename)}"
+      s3_object_key = "#{File.basename(template.filename)}"
```
kaushik, line 211: @name here is deployment name right?  losing that means multiple deployments will use the same path and conflict with each other.

kaushik (changes requested):

lib/moonshot/stack.rb
kaushik: its even better if the path is bucket/app-name/env-name/filename

lib/moonshot/dynamic_template.rb
```diff
@@ -58,9 +58,38 @@ def validate_destination_exists
     end
 
     def validate_template(template)
-      Aws::CloudFormation::Client.new.validate_template(
-        template_body: template
-      )
+      if template.bytesize > 50_000 # Leave some margin from the 51,200 limit
```
kaushik, line 61: @rutuja810 checking if config.template_s3_bucket is set is a better condition here since s3 is used only when this config is set.

PR #338, opened 2026-04-10
CPD-13207: Add a retry of 60m for release download.

rutuja810:
Change: minor
Purpose: maintenance

**Motivation**

Fixes #NNN

**Proposed changes**

**Alternatives considered**

**Testing steps**

### DIFF
--- a/lib/moonshot/artifact_repository/s3_bucket_via_github_releases.rb
+++ b/lib/moonshot/artifact_repository/s3_bucket_via_github_releases.rb
@@ -113,13 +113,18 @@ module Moonshot::ArtifactRepository
     #
     # @param version [String] Version to be downloaded
     # @param [String] Build file downloaded.
-    # @raise [RuntimeError] If the file fails to download correctly after 3
-    #                       attempts.
+    # @raise [RuntimeError] If the file fails to download correctly within 60 minutes.
     def download_from_github(version)
       file_pattern = "*#{version}*.tar.gz"
       attempts = 0
+      retry_opts = {
+        tries: 360, # up to 360 x 10s = 60 minutes
+        max_elapsed_time: 60 * 60, # 60 minutes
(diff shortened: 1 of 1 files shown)


lib/moonshot/artifact_repository/s3_bucket_via_github_releases.rb
```diff
@@ -113,13 +113,18 @@ def upload_to_s3(file, key)
...
-    #                       attempts.
+    # @raise [RuntimeError] If the file fails to download correctly within 60 minutes.
     def download_from_github(version)
       file_pattern = "*#{version}*.tar.gz"
       attempts = 0
-
-      Retriable.retriable on: RuntimeError do
+      retry_opts = {
+        tries: 360, # up to 360 x 10s = 60 minutes
+        max_elapsed_time: 60 * 60, # 60 minutes
```
kaushik, line 122: 60mins just to download a file form github? isnit it a bit excessive?

rutuja810: it will retry for 1h, as pipeline sometime takes 32-33 min to finish.

marciotoshio: Isn't your pipeline that's taking longer? I think this should be configurable and you can define the retry properties in your pipeline, I'd not like mine have a such long retry if I don't need it 😉

rutuja810: Hi Friend :)
We added retry on our pipeline here https://github.com/acquia/cloud-moonshot-plugins/pull/53/changes.
When release tag is cut it's using same sha as commit where pipeline has already passed so it is showing pipeline passed before triggering a new.

shrikantbhosaleacquia (approved):

